PR #22, opened 2023-06-06
feat: hash view helper

### LINKED ISSUE #17: Provide on-chain helper function to compute `H(params)`
In order to cancel a Composable CoW order one has to compute `H(params)` and pass this into the `remove` method. Whereas everything else can be fairly easily done using the Safe UI (and potentially a tool to encode calldata), computing this hash is non trivial and could therefore benefit from a helper view function that makes this computable via the Etherscan UI.

mfw78:
This PR:

1. Provides a `H(params)` function on `ComposableCoW`, providing utility for a user to determine the hash from which to cancel their conditional order (if done using a single order).

On merge, this closes #17.

### DIFF
--- a/src/ComposableCoW.sol
+++ b/src/ComposableCoW.sol
@@ -85,7 +85,7 @@ contract ComposableCoW is ISafeSignatureVerifier {
             revert InvalidHandler();
         }
 
-        singleOrders[msg.sender][keccak256(abi.encode(params))] = true;
+        singleOrders[msg.sender][hash(params)] = true;
         if (dispatch) {
             emit ConditionalOrderCreated(msg.sender, params);
         }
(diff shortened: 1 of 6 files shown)


src/ComposableCoW.sol
```diff
@@ -85,7 +85,7 @@ contract ComposableCoW is ISafeSignatureVerifier {
             revert InvalidHandler();
         }
 
-        singleOrders[msg.sender][keccak256(abi.encode(params))] = true;
+        singleOrders[msg.sender][hash(params)] = true;
```
fleupold, line 88: noice!

fleupold (approved):
Looks good to me, thanks for adding this convenience. The PR also includes the changes from the other PR (making it slightly confusing to review) but no biggie!

PR #23, opened 2023-06-07
feat: create with context

mfw78:
This PR:

1. Allows for setting an on-chain context at the time of mining for the merkle root set / single order creation transaction.

### DIFF
--- a/src/BaseConditionalOrder.sol
+++ b/src/BaseConditionalOrder.sol
@@ -25,7 +25,7 @@ abstract contract BaseConditionalOrder is IConditionalOrderGenerator {
         bytes calldata offchainInput,
         GPv2Order.Data calldata
     ) external view override {
-        (GPv2Order.Data memory generatedOrder, ) = getTradeableOrder(owner, sender, ctx, staticInput, offchainInput);
+        (GPv2Order.Data memory generatedOrder,) = getTradeableOrder(owner, sender, ctx, staticInput, offchainInput);
 
         /// @dev Verify that the order is valid and matches the payload.
         if (_hash != GPv2Order.hash(generatedOrder, domainSeparator)) {
@@ -37,12 +37,13 @@ abstract contract BaseConditionalOrder is IConditionalOrderGenerator {
      * @dev Set the visibility of this function to `public` to allow `verify` to call it.
      * @inheritdoc IConditionalOrderGenerator
      */
-    function getTradeableOrder(address owner, address sender, bytes32 ctx, bytes calldata staticInput, bytes calldata offchainInput)
-        public
-        view
-        virtual
-        override
-        returns (GPv2Order.Data memory, IConditionalOrder.Interactions memory);
+    function getTradeableOrder(
+        address owner,
+        address sender,
+        bytes32 ctx,
+        bytes calldata staticInput,
+        bytes calldata offchainInput
+    ) public view virtual override returns (GPv2Order.Data memory, IConditionalOrder.Interactions memory);
 
     /**
      * @inheritdoc IERC165
--- a/src/BaseConditionalOrder.sol
+++ b/src/BaseConditionalOrder.sol
@@ -20,11 +20,12 @@ abstract contract BaseConditionalOrder is IConditionalOrderGenerator {
         address sender,
         bytes32 _hash,
         bytes32 domainSeparator,
+        bytes32 ctx,
         bytes calldata staticInput,
         bytes calldata offchainInput,
         GPv2Order.Data calldata
     ) external view override {
-        GPv2Order.Data memory generatedOrder = getTradeableOrder(owner, sender, staticInput, offchainInput);
+        GPv2Order.Data memory generatedOrder = getTradeableOrder(owner, sender, ctx, staticInput, offchainInput);
 
         /// @dev Verify that the order is valid and matches the payload.
         if (_hash != GPv2Order.hash(generatedOrder, domainSeparator)) {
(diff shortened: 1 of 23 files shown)


src/interfaces/IConditionalOrder.sol
```diff
@@ -30,6 +31,7 @@ interface IConditionalOrder {
      * @param sender the `msg.sender` of the transaction
      * @param _hash the hash of the order
      * @param domainSeparator the domain separator used to sign the order
+     * @param ctx the context of the order (bytes32(0) if a merkle tree is used, otherwise H(params))
```
fleupold, line 34: Maybe specify the intent here that this value can be used to look up order specific storage in composable cow?

fleupold (dismissed):
Looks good to me. Are you also going to check in a `CurrrentBlockTimestampFactory`?

src/interfaces/IConditionalOrder.sol
mfw78: Done

test/ComposableCoW.twap.t.sol
```diff
@@ -225,6 +231,40 @@ contract ComposableCoWTwapTest is BaseComposableCoWTest {
         twap.getTradeableOrder(address(0), address(0), bytes32(0), abi.encode(o), bytes(""));
     }
 
+    function test_getTradeableOrder_e2e_fuzz_WithContext(uint32 _blockTimestamp, uint256 currentTime) public {
+        // guard against overflows
+        vm.assume(_blockTimestamp < type(uint32).max);
+        vm.assume(currentTime < type(uint32).max);
+        // guard against revert before start
+        vm.assume(_blockTimestamp < currentTime);
```
fleupold, line 239: Not necessarily as a fuzz test, but might be nice to have one test case showing that blockTimestamp is actually used a start time (technically startTime 0 would have worked in the old codebase wouldn't it?)

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2FHw8vYF4DNRCKY%2Fgiphy.gif%3Fcid%3D2972a2b15d1d81cf6d6d7963363a4f7b%26rid%3Dgiphy.gif&token=shipit" data-gitmeme-token="shipit"></a>

test/ComposableCoW.twap.t.sol
mfw78: Absolute fair - incoming!

src/ComposableCoW.sol
```diff
@@ -91,6 +113,28 @@ contract ComposableCoW is ISafeSignatureVerifier {
...
+     * @param data Implementation specific off-chain data
+     * @param dispatch Whether to dispatch the `ConditionalOrderCreated` event
+     */
+    function createWithContext(
+        IConditionalOrder.ConditionalOrderParams calldata params,
+        IValueFactory factory,
+        bytes calldata data,
+        bool dispatch
+    ) external {
+        create(params, dispatch);
```
anxolin, line 129: not sure if there's gas optimizations possible here, or if it's not even worth caring, but I just noted that we hash the params twice (once inside the good old `create` and the other to store in the `cabinet`

src/ComposableCoW.sol
```diff
@@ -130,19 +174,26 @@ contract ComposableCoW is ISafeSignatureVerifier {
         PayloadStruct memory _payload = abi.decode(payload, (PayloadStruct));
 
         // Check if the order is authorised
-        _auth(address(safe), _payload.params, _payload.proof);
+        bytes32 ctx = _auth(address(safe), _payload.params, _payload.proof);
```
anxolin, line 177: I didn't fully get what is `ctx` for. From what i can tell, is returning the hash that identifies the order, and only for `singleOrders`. But this is used later for `getTradeableOrder`. isn't it relevant for the Merkle root flow?
I guess I'm missing something

src/ComposableCoW.sol
```diff
@@ -229,34 +280,43 @@ contract ComposableCoW is ISafeSignatureVerifier {
...
 
-        if (!(proof.length != 0 || singleOrders[owner][hash(params)])) {
-            revert SingleOrderNotAuthed();
+            // Check if the proof is valid
+            if (!MerkleProof.verify(proof, roots[owner], leaf)) {
+                revert ProofNotAuthed();
+            }
+        } else {
+            // Check if the order is authorised
+            ctx = hash(params);
```
anxolin, line 295: just double-ckecing if this should be only in the else, or should be done outside of the IF

src/types/twap/TWAP.sol
```diff
@@ -30,7 +38,16 @@ contract TWAP is BaseConditionalOrder {
...
          * settled once.
          */
-        order = TWAPOrder.orderFor(abi.decode(staticInput, (TWAPOrder.Data)));
+        TWAPOrder.Data memory twap = abi.decode(staticInput, (TWAPOrder.Data));
+
+        /**
+         * @dev If `twap.t0` is set to 0, then get the start time from the context.
+         */
+        if (twap.t0 == 0) {
+            twap.t0 = uint256(composableCow.cabinet(owner, ctx));
```
anxolin, line 47: ❤️

anxolin (commented):
👍 Great changes!

src/ComposableCoW.sol
mfw78: Yep, it should only be done in the `else`. This takes advantage of the `ctx` return parameter already being initialised to the `bytes32(0)` (zero) value, which is the case to return when considering a Merkle flow.

src/BaseConditionalOrder.sol
```diff
@@ -20,11 +20,12 @@ abstract contract BaseConditionalOrder is IConditionalOrderGenerator {
         address sender,
         bytes32 _hash,
         bytes32 domainSeparator,
+        bytes32 ctx,
```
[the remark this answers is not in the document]
mfw78: Sure. Given the `@inheritdoc IConditionalOrder`, this picks up the natdoc from `IConditionalOrder.verify` which contains the `ctx` and other parameter documentation. :+1:

PR #29, opened 2023-08-01
fix: ackee audit

mfw78:
This is a WIP PR. Do not merge yet.

This PR:

1. Applies fixes, with commits in the PR tagged with the ackee audit finding that they fix.
2. Applies additional fixes around consistency (comments, a 'utils' library for common functions in conditional orders).
3. Adds in `appData` to all conditional order types for consistency.
4. Generalises the deployer so that the `salt` for `CREATE2` is configurable via environment variable.

**Todo**:

- [ ]  ~~Oracle data checking in `StopLoss`~~ Tracked in #30


script/deploy_ProdStack.s.sol
```diff
@@ -19,19 +19,20 @@ contract DeployProdStack is Script {
     function run() external {
         uint256 deployerPrivateKey = vm.envUint("PRIVATE_KEY");
         address settlement = vm.envAddress("SETTLEMENT");
+        bytes32 salt = vm.envBytes32("SALT");
 
         vm.startBroadcast(deployerPrivateKey);
 
         // Deploy ExtensibleFallbackHandler
-        new ExtensibleFallbackHandler{salt: "v1"}();
+        new ExtensibleFallbackHandler{salt: salt}();
```
fleupold, line 27: By having the salt hardcoded and checked into code, we can reproduce the builds if needed later (e.g. on other chains). If we pass it as an env, do we need to think about a way to document the used salt (or can it be easily recovered from the deployment tx)?

src/types/StopLoss.sol
```diff
@@ -44,33 +44,31 @@ contract StopLoss is BaseConditionalOrder {
...
+    function getTradeableOrder(address, address, bytes32, bytes calldata staticInput, bytes calldata)
+        public
+        view
+        override
+        returns (GPv2Order.Data memory order)
+    {
         Data memory data = abi.decode(staticInput, (Data));
-        (, int256 latestSellPrice, , , ) = data.sellTokenPriceOracle.latestRoundData();
-        (, int256 latestBuyPrice, , , ) = data.buyTokenPriceOracle.latestRoundData();
+        (, int256 latestSellPrice,,,) = data.sellTokenPriceOracle.latestRoundData();
```
fleupold, line 54: Make sense to do this in a separate PR as it is more involved, but I assume they suggest to use check that the oracle's decimals view method returns the same number for both and check that updated at is not too far in the past (not sure if we should allow the user to specify this though)

test/ComposableCoW.base.t.sol
```diff
@@ -149,10 +149,16 @@ contract BaseComposableCoWTest is Base, Merkle {
...
+        vm.startPrank(owner);
         bytes32 orderHash = keccak256(abi.encode(params));
+        bytes32 ctx = composableCow.cabinet(owner, orderHash);
         composableCow.remove(orderHash);
         assertEq(composableCow.singleOrders(owner, orderHash), false);
+        if (ctx != bytes32(0)) {
+            // ensure that the context was cleared
+            assertEq(composableCow.cabinet(owner, orderHash), bytes32(0));
+        }
+        vm.stopPrank();
```
fleupold, line 161: Why do we need to start/stop prank?

test/ComposableCoW.twap.t.sol
```diff
@@ -322,6 +322,10 @@ contract ComposableCoWTwapTest is BaseComposableCoWTest {
                 GPv2Order.hash(part, settlement.domainSeparator()), signature
             ) == ERC1271.isValidSignature.selector
         );
+
+        // Now remove the order
+        vm.prank(address(safe1));
+        composableCow.remove(keccak256(abi.encode(params)));
```
fleupold, line 328: do we want to assert anything afterwards or are we simply fine with it not reverting (seems weird to test an action without checking a post-assumption)?

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2FdaxVu1mZErt2qKYdtX%2Fgiphy.gif&token=fix" data-gitmeme-token="fix"></a>

test/ComposableCoW.base.t.sol
mfw78: Ahh, this is because the `composableCow.cabinet` call took the `vm.prank()`. Instead I'll move the `vm.prank()` down to precede `composableCoW.remove` and remove the `vm.stopPrank()`

src/types/StopLoss.sol
mfw78: Yes, I'll put this into a separate PR. tl;dr I'm working on actually generalising the strike price such that it's sell token / buy token in ERC20 decimals, and normalising the returned quote from the oracles.

script/deploy_ProdStack.s.sol
mfw78: Fair point. I'll revert this one. I was hoping that it was easy to recover from the broadcast JSON but I can't see it easily.

test/ComposableCoW.twap.t.sol
mfw78: I think from memory there were assertions in another test suite referencing this, but in lieu of that, I've added assertions afterwards to stick to conventions.

fleupold (approved):
Looks great!

PR #33, opened 2023-08-02
Parameterised Errors

### LINKED ISSUE #32: feat: parameterised errors
**Problem**
In the course of the ackee audit, some project consistency items have been higlighted. Amongst these are the use of `error IConditionalOrder.OrderNotValid()` for all conditions when the order isn't valid.

As orders may have complex / not easily rationalised logic, this creates onerous demand on the developer when debugging.

**Solution**

Change `OrderNotValid` to take a `bytes4` parameter. This `bytes4` parameter is itself an `error` selector native to the conditional order type that has been implemented.

mfw78:
This PR:

1. Introduces parameterised errors to help with debugging consistency across order types. Closes #32

### DIFF
--- a/src/interfaces/IConditionalOrder.sol
+++ b/src/interfaces/IConditionalOrder.sol
@@ -11,8 +11,8 @@ import {IERC165} from "safe/interfaces/IERC165.sol";
  */
 interface IConditionalOrder {
     /// @dev This error is returned by the `getTradeableOrder` function if the order condition is not met.
-    ///      A parameter of `bytes4` type is included to allow the caller to specify the reason for the failure.
-    error OrderNotValid(bytes4);
+    ///      A parameter of `string` type is included to allow the caller to specify the reason for the failure.
+    error OrderNotValid(string);
 
     /**
      * @dev This struct is used to uniquely identify a conditional order for an owner.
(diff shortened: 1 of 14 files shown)


src/interfaces/IConditionalOrder.sol
```diff
@@ -11,7 +11,8 @@ import {IERC165} from "safe/interfaces/IERC165.sol";
  */
 interface IConditionalOrder {
     /// @dev This error is returned by the `getTradeableOrder` function if the order condition is not met.
-    error OrderNotValid();
+    ///      A parameter of `bytes4` type is included to allow the caller to specify the reason for the failure.
+    error OrderNotValid(bytes4);
```
fleupold, line 15: Feel free to ignore, but given this is only called in view functions do we care about the gas efficiency? How would we easily map a bytes4 to an error type when debugging (do we need a table for that)? 

Wouldn't it be more convenient to just log a string?

fleupold (approved):
Looks good, just want to make sure it's actually easy to debug a revert reason if all we have is a bytes4 identifier.

src/interfaces/IConditionalOrder.sol
mfw78: This is fair. I was both optimising for gas and also after a way to ensure strong typings on error messages so that typos in tests are discouraged. Instead I broke out the error messages as constants to eliminate the typo problem. Best of both worlds!

fleupold (approved):
Looks great!

PR #34, opened 2023-08-03
fix: ackee oracle price normalisation

mfw78:
This PR:

1. Removes any notion of the ERC20 decimal being used for price comparison, and instead relies on a strike price with 18 decimal places.
2. Moves the scalePrice function to util libs as this may be useful across future oracle-related endeavours.


Alcala1234 (commented):
aecb08866bc3064abe09a5df087d0aed6deab1cd

PR #36, opened 2023-08-08
Do not fail on intermediate errors, but fail at the end if there was …

anxolin:
This PR tries to address some concerns around errors. 

## Hiden errors
Right now, if there's an error processing some logs, or posting the order, Tenderly will still show you that it succeeded.

This is because we catch and handle the errors, so you need to get inside each execution and check the logs in order to see if something is wrong.

## Change applied
This PR will still log every issue, and continue processing until the end, but if there was at least one error, we will throw it at the end of the execution so Tenderly knows that something went badly.

**BEFORE:**
image

**AFTER**
image

This will make it easier to spot the executions that failed

## Not included
Some of these errors could be handled, and shouldn't throw, for example, DuplicatedOrder should be a warning, and not an error (the order is in the system):
image

For now, this PR just want to add more visibility when an error occurs.

### DIFF
--- a/actions/register.ts
+++ b/actions/register.ts
@@ -4,11 +4,17 @@ import {
   Event,
   TransactionEvent,
   Storage,
+  Log,
 } from "@tenderly/actions";
 import { BytesLike, ethers } from "ethers";
 
-import type { ComposableCoW, IConditionalOrder } from "./types/ComposableCoW";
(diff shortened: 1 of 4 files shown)


mfw78 (approved):
LGTM :+1:

PR #40, opened 2023-08-11
Improve logs and dont fail if it is a duplicated order

anxolin:
This PR do a series of improvements to have better visibility on our indexing so we can:
- be notified in case of an error
- can debug the errors easier

## Dont show as error when Duplicated Orders
When an order is duplicated means its already in the orderbook. Therefore, we shouldn't mark the indexing as an error. This PR shows a warning but continues. The order will be registered and monitored in the same way as if the registration in the order book was successful.

image

# Notify errors in slack
This PR will notify any error in slack

Its not an ideal solution, but a patch until we have the Valory implementation in place

image

## Show more context in the logs
Now logs show more context, like the ethereum transaction where the logs were found in the first place.

I found this is useful when debugging issues. 

I made it so its part of the model, so now is persisted and logged. 

## Overengineered solution to fight some logging problems with Tenderkly logs
Tenderly logs have a limitation of 4K, if you reach it, they will disable them so you will miss a lot of important information.

This PR overrides the console.log so, in case you try to log something bigger (a big JSON) it will try to break the message into several logs. 

The console log is overridden cause i suspected that one internal library was logging something big, and that was breaking Tendrly. This override helped a lot, and now the logs are more clear, however, i still see from time to time we reach the limitation for some reason I don't understand

### DIFF
--- a/actions/register.ts
+++ b/actions/register.ts
@@ -14,7 +14,14 @@ import type {
   IConditionalOrder,
 } from "./types/ComposableCoW";
 import { ComposableCoW__factory } from "./types/factories/ComposableCoW__factory";
-import { logger, writeRegistry } from "./utils";
+import { handleExecutionError, init, writeRegistry } from "./utils";
+
+// Standardise the storage key
+const LAST_NOTIFIED_ERROR_STORAGE_KEY = "LAST_NOTIFIED_ERROR";
+
+export const getOrdersStorageKey = (network: string): string => {
+  return `CONDITIONAL_ORDER_REGISTRY_${network}`;
+};
 
 /**
  * Listens to these events on the `ComposableCoW` contract:
@@ -24,30 +31,30 @@ import { logger, writeRegistry } from "./utils";
  * @param event transaction event
  */
 export const addContract: ActionFn = async (context: Context, event: Event) => {
+  return _addContract(context, event).catch(handleExecutionError);
+};
+
+const _addContract: ActionFn = async (context: Context, event: Event) => {
   const transactionEvent = event as TransactionEvent;
   const tx = transactionEvent.hash;
   const composableCow = ComposableCoW__factory.createInterface();
-
-  // Load the registry
-  const registry = await Registry.load(context, transactionEvent.network);
+  const { registry } = await init(transactionEvent.network, context);
 
   // Process the logs
   let hasErrors = false;
   transactionEvent.logs.forEach((log) => {
-    const { error } = _addContract(tx, log, composableCow, registry);
+    const { error } = _registerNewOrder(tx, log, composableCow, registry);
     hasErrors ||= error;
   });
 
-  hasErrors ||= !(await writeRegistry(registry));
-  // Notify error
+  hasErrors ||= !(await writeRegistry());
+  // Throw execution error if there was at least one error
   if (hasErrors) {
-    // TODO notify error to slack
-
     throw Error("Error adding contract for event" + JSON.stringify(event));
   }
 };
 
-export function _addContract(
+export function _registerNewOrder(
   tx: string,
   log: Log,
   composableCow: ComposableCoWInterface,
@@ -105,9 +112,9 @@ export function _addContract(
       }
     }
   } catch (error) {
-    logger.error(
+    console.error(
       "[addContract] Error handling ConditionalOrderCreated/MerkleRootSet event" +
-        error // extractErrorMessage(error)
+        error
     );
     return { error: true };
   }
@@ -134,8 +141,8 @@ export const add = async (
 ) => {
   if (registry.ownerOrders.has(owner)) {
     const conditionalOrders = registry.ownerOrders.get(owner);
-    logger.log(
-      `[register:add] Adding conditional order ${params} to already existing contract ${owner}`
+    console.log(
+      `[register:add] Adding conditional order ${params} to already existing contract ${owner}. Tx: ${tx}`
     );
     let exists: boolean = false;
     // Iterate over the conditionalOrders to make sure that the params are not already in the registry
@@ -158,8 +165,8 @@ export const add = async (
       });
     }
   } else {
-    logger.log(
-      `[register:add] Adding conditional order ${params} to new contract ${owner}`
+    console.log(
+      `[register:add] Adding conditional order ${params} to new contract ${owner} . Tx: ${tx}`
     );
     registry.ownerOrders.set(
       owner,
@@ -230,12 +237,14 @@ export type ConditionalOrder = {
 };
 
 /**
- * A registry is a map of owners to conditional orders.
+ * Models the state beteween executions.
+ * Contains a map of owners to conditional orders and the last time we sent an error.
  */
 export class Registry {
   ownerOrders: Map<Owner, Set<ConditionalOrder>>;
   storage: Storage;
   network: string;
+  lastNotifiedError: Date | null;
 
   /**
    * Instantiates a registry.
@@ -246,11 +255,13 @@ export class Registry {
   constructor(
     ownerOrders: Map<Owner, Set<ConditionalOrder>>,
     storage: Storage,
-    network: string
+    network: string,
+    lastNotifiedError: Date | null
   ) {
     this.ownerOrders = ownerOrders;
     this.storage = storage;
     this.network = network;
+    this.lastNotifiedError = lastNotifiedError;
   }
 
   /**
@@ -263,17 +274,28 @@ export class Registry {
     context: Context,
     network: string
   ): Promise<Registry> {
-    const str = await context.storage.getStr(storageKey(network));
+    const str = await context.storage.getStr(getOrdersStorageKey(network));
+    const lastNotifiedError = await context.storage
+      .getStr(LAST_NOTIFIED_ERROR_STORAGE_KEY)
+      .then((isoDate) => (isoDate ? new Date(isoDate) : null))
+      .catch(() => null);
+
     if (str === null || str === undefined || str === "") {
       return new Registry(
         new Map<Owner, Set<ConditionalOrder>>(),
         context.storage,
-        network
+        network,
+        lastNotifiedError
       );
     }
 
     const ownerOrders = JSON.parse(str, reviver);
-    return new Registry(ownerOrders, context.storage, network);
+    return new Registry(
+      ownerOrders,
+      context.storage,
+      network,
+      lastNotifiedError
+    );
   }
 
   /**
@@ -281,9 +303,21 @@ export class Registry {
    */
   public async write() {
     await this.storage.putStr(
-      storageKey(this.network),
+      getOrdersStorageKey(this.network),
       JSON.stringify(this.ownerOrders, replacer)
     );
+
+    if (this.lastNotifiedError !== null) {
+      console.log(
+        "this.lastNotifiedError.toISOString()",
+        this.lastNotifiedError,
+        typeof this.lastNotifiedError
+      );
+      await this.storage.putStr(
+        LAST_NOTIFIED_ERROR_STORAGE_KEY,
+        this.lastNotifiedError.toISOString()
+      );
+    }
   }
 }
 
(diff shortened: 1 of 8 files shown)


actions/register.ts
```diff
@@ -153,11 +168,11 @@ export const add = async (
     }
   } else {
     console.log(
-      `[register:add] Adding conditional order ${params} to new contract ${owner}`
+      `[register:add] Adding conditional order ${params} to new contract ${owner} . Tx: ${tx}`
```
anxolin, line 171: it helps a lot to have the TX hash in the logs

actions/register.ts
```diff
@@ -255,27 +276,48 @@ export class Registry {
     context: Context,
     network: string
   ): Promise<Registry> {
-    const str = await context.storage.getStr(storageKey(network));
+    const str = await context.storage.getStr(getOrdersStorageKey(network));
+    const lastNotifiedError = await context.storage
+      .getStr(LAST_NOTIFIED_ERROR_STORAGE_KEY)
+      .then((isoDate) => (isoDate ? new Date(isoDate) : null))
+      .catch(() => null);
```
anxolin, line 283: lastNotified is persisted so we avoid creating too many notifications in slack (not super sofisticated, but a small patch for the time being)

actions/utils.ts
```diff
@@ -4,6 +4,57 @@ import assert = require("assert");
...
+  notificationsEnabled: boolean;
+  context: Context;
+}
+
+let executionContext: ExecutionContext | undefined;
+
+export async function init(
+  network: string,
+  context: Context
+): Promise<ExecutionContext> {
```
anxolin, line 23: convenient new function. Each action will make sure calls it. This way we can have some execution context ready that can be used, for example, to have the Slack client ready  to notify in case of an error

actions/utils.ts
```diff
@@ -85,15 +156,114 @@ function handlePromiseErrors<T>(
     .then(() => true)
     .catch((error) => {
       console.error(errorMessage, error);
-      return true;
+      return false;
```
anxolin, line 159: 🙈 this was a mistake!

actions/utils.ts
```diff
@@ -85,15 +156,114 @@ function handlePromiseErrors<T>(
...
+
+  return true;
+}
+
+var consoleOriginal = {
+  log: console.log,
+  error: console.error,
+  warn: console.warn,
+  debug: console.debug,
+};
```
anxolin, line 185: I have overridden the console, to make sure I can prevent from writing too big messages, which breaks Tenderly

anxolin: At the beginning i was only defining my own logger function, but since i had still issues with Tenderly removing logs, I became suspicious that some library was also logging something and this was breaking Tenderly,

actions/utils.ts
```diff
@@ -85,15 +156,114 @@ function handlePromiseErrors<T>(
...
+      consoleOriginal[level](message);
+
+      // if (level === "error") {
+      //   sendSlack(message);
+      // }
+
+      // // Used to debug the Tender log Limit issues
+      // consoleOriginal[level](
+      //   prefix + "TEST for bigLogText of " + bigLogText.length + " bytes"
+      // );
```
anxolin, line 223: I can remove if it bothers the reviewers. It help me a lot in debugging the issues with the logs being removed. I don't consider the problem 100% solved, so i left it as a comment to quickly uncomment if needed

actions/watch.ts
```diff
@@ -36,34 +43,51 @@ const GPV2SETTLEMENT = "0x9008D19f58AAbD9eD0D60971565AA8510560ab41";
...
+ */
+const _checkForSettlement: ActionFn = async (
+  context: Context,
+  event: Event
 ) => {
   const transactionEvent = event as TransactionEvent;
   const settlement = GPv2Settlement__factory.createInterface();
 
-  const registry = await Registry.load(context, transactionEvent.network);
+  const { registry } = await init(transactionEvent.network, context);
```
anxolin, line 60: now building the registry is not responsibility of the actions itself

They just need to init the executionContext

actions/watch.ts
```diff
@@ -73,22 +97,24 @@ async function _checkForSettlement(
...
-        const conditionalOrders = registry.ownerOrders.get(owner);
+        const conditionalOrders = ownerOrders.get(owner);
         // Iterate over the conditionalOrders and update the status of the orderUid
         conditionalOrders?.forEach((conditionalOrder) => {
           // Check if the orderUid is in the conditionalOrder
           if (conditionalOrder.orders.has(orderUid)) {
             // Update the status of the orderUid to FILLED
-            console.log(`Update order ${orderUid} to status FILLED`);
+            console.log(
+              `Update order ${orderUid} to status FILLED. Settlement Tx: ${tx}`
```
anxolin, line 109: tx hash is useful, now we can track the order to the TX that created the smart order and resulted in creating this orderUid, that now is filled :)

actions/watch.ts
```diff
@@ -320,9 +374,18 @@ async function placeOrder(orderUid: string, order: any, apiUrl: string) {
     const errorMessage = "[placeOrder] Error placing order in API";
     if (error.response) {
       const { status, data } = error.response;
+
+      const { shouldThow } = handleOrderBookError(status, data);
```
anxolin, line 378: For some error we will want to throw, and for some others, we will want to ignore. `handleOrderBookError` will tell us what we should do

actions/utils.ts
```diff
@@ -85,15 +156,114 @@ function handlePromiseErrors<T>(
...
+
+var consoleOriginal = {
+  log: console.log,
+  error: console.error,
+  warn: console.warn,
+  debug: console.debug,
+};
+
+/**
+ * Tenderly has a limit of 4Kb per log message. When you surpas this limit, the log is not printed any more making it super hard to debug anythnig
```
mfw78, line 188:
```suggestion
 * Tenderly has a limit of 4Kb per log message. When you surpass this limit, the log is not printed any more making it super hard to debug anything
```

actions/utils.ts
```diff
@@ -85,15 +156,114 @@ function handlePromiseErrors<T>(
...
+      // }
+
+      // // Used to debug the Tender log Limit issues
+      // consoleOriginal[level](
+      //   prefix + "TEST for bigLogText of " + bigLogText.length + " bytes"
+      // );
+    }
+  };
+
+// Override the log function since some internal libraries might print somnething and breaks Tendrly
```
mfw78, line 227:
```suggestion
// Override the log function since some internal libraries might print something and breaks Tenderly
```

actions/utils.ts
```diff
@@ -85,15 +156,114 @@ function handlePromiseErrors<T>(
...
+      const message =
+        prefix +
+        bigLogText.substring(chartStart, chartStart + TENDERLY_LOG_LIMIT);
+      consoleOriginal[level](message);
+
+      // if (level === "error") {
+      //   sendSlack(message);
+      // }
+
+      // // Used to debug the Tender log Limit issues
```
mfw78, line 220:
```suggestion
      // // Used to debug the Tenderly log Limit issues
```

actions/utils.ts
```diff
@@ -71,6 +122,26 @@ export function formatStatus(status: OrderStatus) {
...
+    consoleOriginal.error("Error sending slack notification", error);
+  }
+
+  // Re-throws the original error
+  throw e;
+}
+
 /**
  * Utility function to handle promise, so they are logged in case of an error. It will return a promise that resolves to true if the promise is successful
  * @param errorMessage message to log in case of an error (together witht he original error)
```
mfw78, line 147:
```suggestion
 * @param errorMessage message to log in case of an error (together with the original error)
```

actions/watch.ts
```diff
@@ -36,34 +43,51 @@ const GPV2SETTLEMENT = "0x9008D19f58AAbD9eD0D60971565AA8510560ab41";
 export const checkForSettlement: ActionFn = async (
   context: Context,
   event: Event
+) => {
+  return _checkForSettlement(context, event).catch(handleExecutionError);
+};
+
+/**
+ * Asyncronous version of checkForSettlement. It will process all the settlements, and will throw an error at the end if there was at least one error
```
mfw78, line 51:
```suggestion
 * Asynchronous version of checkForSettlement. It will process all the settlements, and will throw an error at the end if there was at least one error
```

actions/watch.ts
```diff
@@ -268,11 +315,13 @@ function getOrderUid(network: string, orderToSubmit: Order, owner: string) {
  * @param orders All the orders that are being tracked
  */
 export const printUnfilledOrders = (orders: Map<BytesLike, OrderStatus>) => {
-  console.log(`Unfilled orders (${orders.size}):`);
-  for (const [orderUid, status] of orders.entries()) {
-    if (status === OrderStatus.SUBMITTED) {
-      console.log(orderUid);
-    }
+  const unfilledOrders = Array.from(orders.entries())
+    .filter(([_orderUid, status]) => status === OrderStatus.SUBMITTED)
```
mfw78, line 319:
```suggestion
    .filter(([_orderUid, status]) => status === OrderStatus.SUBMITTED) // as SUBMITTED != FILLED
```

Just a bit of documentation to make it a little more obvious as to why we're filtering on `SUBMITTED`. In hind sight, I should have included.

actions/watch.ts
```diff
@@ -302,7 +355,8 @@ async function placeOrder(orderUid: string, order: any, apiUrl: string) {
     };
 
     // if the api_url doesn't contain localhost, post
```
mfw78, line 357:
```suggestion
    // if the apiUrl doesn't contain localhost, post
```

actions/watch.ts
mfw78, line 378:
```suggestion
      const { shouldThrow } = handleOrderBookError(status, data);
```

actions/watch.ts
```diff
@@ -320,9 +374,18 @@ async function placeOrder(orderUid: string, order: any, apiUrl: string) {
     const errorMessage = "[placeOrder] Error placing order in API";
     if (error.response) {
       const { status, data } = error.response;
+
+      const { shouldThow } = handleOrderBookError(status, data);
+
       // The request was made and the server responded with a status code
       // that falls out of the range of 2xx
-      console.error(`${errorMessage}. Result: ${status}`, data);
+      const log = console[shouldThow ? "error" : "warn"];
```
mfw78, line 382:
```suggestion
      const log = console[shouldThrow ? "error" : "warn"];
```

actions/watch.ts
```diff
@@ -320,9 +374,18 @@ async function placeOrder(orderUid: string, order: any, apiUrl: string) {
...
+      log(`${errorMessage}. Result: ${status}`, data);
+
+      if (!shouldThow) {
```
mfw78, line 385:
```suggestion
      if (!shouldThrow) {
```

actions/watch.ts
```diff
@@ -405,3 +468,11 @@ class ChainContext {
     return new ChainContext(provider, apiUrl(network));
   }
 }
+function handleOrderBookError(status: any, data: any): { shouldThow: boolean } {
```
mfw78, line 471:
```suggestion
function handleOrderBookError(status: any, data: any): { shouldThrow: boolean } {
```

actions/watch.ts
```diff
@@ -405,3 +468,11 @@ class ChainContext {
...
+  if (status === 400 && data?.errorType === "DuplicatedOrder") {
+    // The order is in the OrderBook, all good :)
+    return { shouldThow: false };
```
mfw78, line 474:
```suggestion
    return { shouldThrow: false };
```

actions/watch.ts
```diff
@@ -405,3 +468,11 @@ class ChainContext {
...
+  }
+
+  return { shouldThow: true };
```
mfw78, line 477:
```suggestion
  return { shouldThrow: true };
```

mfw78 (approved):
Looks great! Just minor nits on typos :smiley:

PR #41, opened 2023-08-14
Allow to re-process old blocks from local run

anxolin:
This PR is to make debugging the WatchTower, and even more importantly, to fix some issues in production that might arise (so we can make ONCALL easier)

## Run actions for a specific block

It makes a small modification in the script that runs locally the actions. Now, it allows you tu run it, instead of in "watch mode" (watching for new blocks), you will be able to run it for a specific block.

## On-call
The idea is, in case during ONCALL someone complaints their smart order is not becoming OPEN when it should, we can:
- Ask the user for a TX, check the block number
- Add that block number in our `.env` file (see this PR changes in the example file)
- Run the actions locally

## How can this be used
This will allow to:
- Create an order in the orderbook but for some reason the WatchTowers failed to create it (i.e. they were temporarily down)
- Debug an issue that prevented the WatchTower to create the order (see image below)

image

# More context
I added some comments in the code changes

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -18,7 +18,7 @@ NETWORK=5 # Network to test. 1: Mainnet, 5: Goerli, 100: xDai
 # NODE_USER_100=
 # NODE_PASSWORD_100=
 
-# Optionally, specidy the block number to process locally
+# Optionally, specify the block number to process locally
 #   - useful for debugging or re-apply an old block to make sure the orders are created)
 #   - if not specified, it will run in WATCH mode, subscribing to all new blocks
 #BLOCK_NUMBER=9500767
--- a/.env.example
+++ b/.env.example
@@ -18,6 +18,12 @@ NETWORK=5 # Network to test. 1: Mainnet, 5: Goerli, 100: xDai
 # NODE_USER_100=
 # NODE_PASSWORD_100=
 
+# Optionally, specidy the block number to process locally
+#   - useful for debugging or re-apply an old block to make sure the orders are created)
+#   - if not specified, it will run in WATCH mode, subscribing to all new blocks
+#BLOCK_NUMBER=9500767
+BLOCK_NUMBER=
+
 # Slack
 SLACK_WEBHOOK_URL=
 NOTIFICATIONS_ENABLED=false
--- a/README.md
+++ b/README.md
@@ -329,7 +329,7 @@ yarn build:actions
 
 # Run actions locally
 #   - It will start watching and processing new blocks
-#   - As a result, new Composable Cow orders will be discovered and posted to the Orederbook API
+#   - As a result, new Composable Cow orders will be discovered and posted to the OrderBook API
 yarn start:actions
 
 # You can re-process an old block by:
--- a/README.md
+++ b/README.md
@@ -308,7 +308,9 @@ For local integration testing, including the use of [Tenderly Actions](#Tenderly
    SAFE="address here" forge script script/submit_SingleOrder.s.sol:SubmitSingleOrder --rpc-url http://127.0.0.1:8545 --broadcast
    ```
 
-#### Local test for Tenderly Web3 Actions
+#### Run Tenderly Web3 Actions locally
+
+> Useful for debugging locally the actions. Also could be used to create an order for an old block in case there was a failure of WatchTowers indexing it.
 
 Make sure you setup the environment (so you have your own `.env` file).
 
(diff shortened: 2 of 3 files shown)


actions/test/run_local.ts
```diff
@@ -37,13 +37,27 @@ const main = async () => {
...
+  const onNewBlock = async (blockNumber: number) => {
     try {
       processBlock(provider, blockNumber, chainId, testRuntime);
     } catch (error) {
       console.error("[run_local] Error in processBlock", error);
     }
-  });
+  };
+
+  if (process.env.BLOCK_NUMBER) {
```
anxolin, line 51: new env to run the actions only for a specified block (instead of running in WATCH mode)

actions/test/run_local.ts
```diff
@@ -111,6 +109,23 @@ async function processBlock(
...
+
+  // run action
+  await testRuntime
+    .execute(checkForAndPlaceOrder, testBlockEvent)
+    .catch((e) => {
+      console.log(
+        "[run_local] Error in checkForAndPlaceOrder processing block",
+        e
+      );
+    });
```
anxolin, line 128: I moved the order of the execution of the actions.

Order matters here, because `addContract` will register the order, and `checkForAndPlaceOrder` will find the order persisted and will create it in the orderbook. 

This part was making the script to add the order in the orderbook for NEXT BLOCK not in the current one where it was defined.

.env.example
```diff
@@ -18,6 +18,12 @@ NETWORK=5 # Network to test. 1: Mainnet, 5: Goerli, 100: xDai
 # NODE_USER_100=
 # NODE_PASSWORD_100=
 
+# Optionally, specidy the block number to process locally
```
mfw78, line 21:
```suggestion
# Optionally, specify the block number to process locally
```

README.md
```diff
@@ -325,6 +327,13 @@ NODE_PASSWORD_100=optionally-provide-password-if-auth-is-required
 # Build Actions
 yarn build:actions
 
-# Run actions locally, so it starts to checking for new blocks, and executing the actions to create Composable Cow orders)
+# Run actions locally
+#   - It will starts watching and processing new blocks
```
mfw78, line 331:
```suggestion
#   - It will start watching and processing new blocks
```

README.md
```diff
@@ -325,6 +327,13 @@ NODE_PASSWORD_100=optionally-provide-password-if-auth-is-required
...
+#   - As a result, new Composable Cow orders will be discovered and posted to the Orederbook API
```
mfw78, line 332:
```suggestion
#   - As a result, new Composable Cow orders will be discovered and posted to the OrderBook API
```

actions/test/run_local.ts
```diff
@@ -37,13 +37,27 @@ const main = async () => {
   const provider = await getProvider(testRuntime.context, network);
   const { chainId } = await provider.getNetwork();
 
-  provider.on("block", async (blockNumber) => {
+  // Provides a way to process an specific block p
```
mfw78, line 40:
```suggestion
  // Provides a way to process a specific block p
```

actions/test/run_local.ts
```diff
@@ -37,13 +37,27 @@ const main = async () => {
...
       processBlock(provider, blockNumber, chainId, testRuntime);
     } catch (error) {
       console.error("[run_local] Error in processBlock", error);
     }
-  });
+  };
+
+  if (process.env.BLOCK_NUMBER) {
+    const blockNumber = Number(process.env.BLOCK_NUMBER);
+    console.log(`[run_local] Processing an specicfic block ${blockNumber}...`);
```
mfw78, line 53:
```suggestion
    console.log(`[run_local] Processing specific block ${blockNumber}...`);
```

actions/test/run_local.ts
mfw78: Nice catch! Yeah, this is an obvious one :see_no_evil: my bad.

mfw78 (dismissed):
Looks good, only minor nits.

FYI, when replaying blocks, this is great, _however_, this may become an issue on chains with a shorter block time (such as Gnosis Chain). Notes may want to be added to the `README.md` for the one-block that if the block being targeted exceeds the state retention of the RPC, use of an archive node is required.

This will simplify ONCALL so they don't get stumped when this as well fails in addition to what was a previous fail case in something unrelated.

mfw78 (approved):

PR #43, opened 2023-08-16
Add sentry

anxolin:
# Summary

Tenderly Actions logs can't be trusted 100%, they currently have an issue which make us have some blind spots on what is happening. 

for this reason i integrated the WatchTower with tenderly, so we can debug better cases we encounter in production:

image

Sentry project: https://cowprotocol.sentry.io/projects/watch-tower-tenderly/?project=4505714328928256

# To Test
- Login in Sentry
- Observe the test error i manually created

PR #44, opened 2023-08-18
Refactor actions and move model

anxolin:
This PR does a small refacrtor to reduce the complexity of the files. 

The main refactor consist of:
- Each action is in each own file (with the name of the action). So 3 actions, 3 files. Before `watch.rs` had `addContract`, and `checkForAndPlaceOrder`, `checkForSettlement`
- The actions don't have dependencies between them (before there were some models and types defined in the action files). Now action will import some shared utils and models from `utils.ts`  and `model.ts` respectively
- All the type definitions and classes have been moved to `model.ts`

## Additionally
- Cleanup some dead-code pointed out in #43 and addressed in this one
- Fixed an issue writing the registry ([see issue here](https://github.com/cowprotocol/composable-cow/pull/44/files#diff-19bd4b5c3ee627f02c69cc3e706f851803fb2f3a484b50fca1ec03ac788427ebR195)). Now `_handleOrderBookError` is asynchronous. I notice an error happening because we were not waiting for a promise that writes in the registry.

## How to review this PR
Apologies for the number of changes, but its really moving things around only.
The best way to review in my opinion is to checkout the code and check the 3 action files and the model

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ node_modules/
 /actions/types/
 /cli/artifacts
 /cli/types
+yarn.lock
--- a/actions/addContract.ts
+++ b/actions/addContract.ts
@@ -3,7 +3,6 @@ import {
   Context,
   Event,
   TransactionEvent,
-  Storage,
   Log,
 } from "@tenderly/actions";
 import { BytesLike, ethers } from "ethers";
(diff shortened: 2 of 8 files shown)


mfw78 (approved):
LGTM! 🚀

PR #46, opened 2023-08-18
Improve error handling on getTradeableOrderWithSignature CALL simulation

anxolin:
Most of the erros happen during `getTradeableOrderWithSignature` `CALL` simulation.

There's many things that can go wrong here.

This error tries to make it more explicit the error handling, so it signals igiven the error, we should re-attempt later. Also, if this is expected or just an un-expeted errors. 

Expected errors won't lead to the TENDERLY action to fail. 
Un-expected will.

If the error handler decides that we shouldn't re-attempt later (its a final error), then it will signal that the order should be deleted. 

## Special mention
I changed slightly the implementation, @mfw78 pls double check if it makes sense now. 

Before, we were deleting the order in case of unforeseen or un-recognised errors. This could be for example because of a temporal network failure or rate limiting or sth, so we should definitely re-attempt later.

see https://github.com/cowprotocol/composable-cow/pull/46/files#diff-8e2dc90f4a6ebdf01f8603707f37f7794972fdf1827102bfce89497804f72837R398

## Edit 
Analysing [this case](https://cowprotocol.sentry.io/issues/4404087467/?project=4505714328928256&query=is%3Aunresolved+level%3Aerror&referrer=issue-stream&sort=date&statsPeriod=14d&stream_index=0 ), i changed this PR to make less noise when 

This order was created, and there was an AUTH, but it was revoked after doing a cancelation.

Its fine to delete the order, and I don't think we should make so much noise in these cases: 33e80fa3e71d4f7e5b5326e3dcccfb4c1e75b1ff

### DIFF
--- a/actions/checkForAndPlaceOrder.ts
+++ b/actions/checkForAndPlaceOrder.ts
@@ -378,7 +378,7 @@ function _handleGetTradableOrderCall(
         };
       case "SingleOrderNotAuthed":
         // If there's no autorization we delete the order
-        // - One reason could be, because the user CACTIONed the order
+        // - One reason could be, because the user CANCELLED the order
         // - for now it doesn't support more advance cases where the order is auth during a pre-interaction
 
         console.info(
@@ -390,7 +390,7 @@ function _handleGetTradableOrderCall(
         };
       case "ProofNotAuthed":
         // If there's no autorization we delete the order
-        // - One reason could be, because the user CACTIONed the order
+        // - One reason could be, because the user CANCELLED the order
         // - for now it doesn't support more advance cases where the order is auth during a pre-interaction
 
         console.info(
--- a/actions/checkForAndPlaceOrder.ts
+++ b/actions/checkForAndPlaceOrder.ts
@@ -378,20 +378,28 @@ function _handleGetTradableOrderCall(
         };
       case "SingleOrderNotAuthed":
         // If there's no autorization we delete the order
-        // for now it doesn't support more advance cases where the order is auth during a pre-interaction
+        // - One reason could be, because the user CACTIONed the order
+        // - for now it doesn't support more advance cases where the order is auth during a pre-interaction
 
-        console.error(
+        console.info(
           `${errorMessagePrefix}: Single order on safe ${owner} not authed. Deleting order...`
         );
-        return { result: CallResult.Failed, deleteConditionalOrder: true };
+        return {
+          result: CallResult.FailedButIsExpected,
+          deleteConditionalOrder: true,
+        };
       case "ProofNotAuthed":
         // If there's no autorization we delete the order
-        // for now it doesn't support more advance cases where the order is auth during a pre-interaction
+        // - One reason could be, because the user CACTIONed the order
+        // - for now it doesn't support more advance cases where the order is auth during a pre-interaction
 
-        console.error(
+        console.info(
           `${errorMessagePrefix}: Proof on safe ${owner} not authed. Deleting order...`
         );
-        return { result: CallResult.Failed, deleteConditionalOrder: true };
+        return {
+          result: CallResult.FailedButIsExpected,
+          deleteConditionalOrder: true,
+        };
     }
 
     console.error(errorMessagePrefix + " for unexpected reasons", error);
--- a/actions/checkForAndPlaceOrder.ts
+++ b/actions/checkForAndPlaceOrder.ts
@@ -116,18 +116,22 @@ async function _processConditionalOrder(
 ): Promise<{ deleteConditionalOrder: boolean; error: boolean }> {
   let error = false;
   try {
-    const result = await _getTradeableOrderWithSignature(
+    const tradableOrderResult = await _getTradeableOrderWithSignature(
       owner,
       conditionalOrder,
       contract
     );
 
-    if (!result.success) {
-      // The simulation failed, this only means the order is not ready to be placed in the orderbook. Will check again in the next block
-      return { error: false, deleteConditionalOrder: false };
+    // Return early if the simulation fails
+    if (tradableOrderResult.result != CallResult.Success) {
+      const { deleteConditionalOrder, result } = tradableOrderResult;
+      return {
+        error: result !== CallResult.FailedButIsExpected, // If we expected the call to fail, then we don't consider it an error
+        deleteConditionalOrder,
+      };
     }
 
-    const { order, signature } = result.data;
+    const { order, signature } = tradableOrderResult.data;
 
     const orderToSubmit: Order = {
       ...order,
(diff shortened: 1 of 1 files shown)


actions/checkForAndPlaceOrder.ts
```diff
@@ -115,12 +116,23 @@ async function _processConditionalOrder(
 ): Promise<{ deleteConditionalOrder: boolean; error: boolean }> {
   let error = false;
   try {
-    const { order, signature } = await _getTradeableOrderWithSignature(
+    const tradableOrderResult = await _getTradeableOrderWithSignature(
```
mfw78, line 119:
```suggestion
    const tradeableOrderResult = await _getTradeableOrderWithSignature(
```

actions/checkForAndPlaceOrder.ts
```diff
@@ -115,12 +116,23 @@ async function _processConditionalOrder(
...
       owner,
       conditionalOrder,
       contract
     );
 
+    // Return early if the simulation fails
+    if (tradableOrderResult.result != CallResult.Success) {
```
mfw78, line 126:
```suggestion
    if (tradeableOrderResult.result != CallResult.Success) {
```

actions/checkForAndPlaceOrder.ts
```diff
@@ -115,12 +116,23 @@ async function _processConditionalOrder(
...
+      const { deleteConditionalOrder, result } = tradableOrderResult;
```
mfw78, line 127:
```suggestion
      const { deleteConditionalOrder, result } = tradeableOrderResult;
```

actions/checkForAndPlaceOrder.ts
```diff
@@ -115,12 +116,23 @@ async function _processConditionalOrder(
...
+      return {
+        error: result !== CallResult.FailedButIsExpected, // If we expected the call to fail, then we don't consider it an error
+        deleteConditionalOrder,
+      };
+    }
+
+    const { order, signature } = tradableOrderResult.data;
```
mfw78, line 134:
```suggestion
    const { order, signature } = tradeableOrderResult.data;
```

actions/checkForAndPlaceOrder.ts
```diff
@@ -299,11 +290,35 @@ function _handleOrderBookError(
   return { shouldThrow: true };
 }
 
+type GetTradeableOrderWithSignatureResult =
+  | GetTradeableOrderWithSignatureSuccess
```
mfw78, line 294: Was there meant to be another element in the type that precedes this, or is this an extraneous 'or' operator?

actions/checkForAndPlaceOrder.ts
```diff
@@ -319,20 +334,82 @@ async function _getTradeableOrderWithSignature(
...
+    switch (error.errorName) {
+      case "OrderNotValid":
+        // The conditional order has not expired, or been cancelled, but the order is not valid
+        // For example, with TWAPs, this may be after `span` seconds have passed in the epoch.
+        return {
+          result: CallResult.FailedButIsExpected,
+          deleteConditionalOrder: false,
+        };
+      case "SingleOrderNotAuthed":
+        // If there's no autorization we delete the order
```
mfw78, line 380:
```suggestion
        // If there's no authorization we delete the order
```

actions/checkForAndPlaceOrder.ts
```diff
@@ -319,20 +334,82 @@ async function _getTradeableOrderWithSignature(
...
+        // - One reason could be, because the user CANCELLED the order
+        // - for now it doesn't support more advance cases where the order is auth during a pre-interaction
```
mfw78, line 382:
```suggestion
        // - for now it doesn't support more advanced cases where the order is auth during a pre-interaction
```

actions/checkForAndPlaceOrder.ts
```diff
@@ -319,20 +334,82 @@ async function _getTradeableOrderWithSignature(
...
+
+        console.info(
+          `${errorMessagePrefix}: Single order on safe ${owner} not authed. Deleting order...`
+        );
+        return {
+          result: CallResult.FailedButIsExpected,
+          deleteConditionalOrder: true,
+        };
+      case "ProofNotAuthed":
+        // If there's no autorization we delete the order
```
mfw78, line 392:
```suggestion
        // If there's no authorization we delete the order
```

actions/checkForAndPlaceOrder.ts
```diff
@@ -319,20 +334,82 @@ async function _getTradeableOrderWithSignature(
...
+        // - One reason could be, because the user CANCELLED the order
+        // - for now it doesn't support more advance cases where the order is auth during a pre-interaction
```
mfw78, line 394:
```suggestion
        // - for now it doesn't support more advanced cases where the order is auth during a pre-interaction
```

actions/checkForAndPlaceOrder.ts
```diff
@@ -319,20 +334,82 @@ async function _getTradeableOrderWithSignature(
...
+    }
+
+    console.error(errorMessagePrefix + " for unexpected reasons", error);
+    // If we don't know the reason, is better to not delete the order
+    return { result: CallResult.Failed, deleteConditionalOrder: false };
+  }
+
+  console.error("[getTradeableOrderWithSignature] Unexpected error", error);
+  // If we don't know the reason, is better to not delete the order
+  return { result: CallResult.Failed, deleteConditionalOrder: false };
```
mfw78, line 412: One of the main failure conditions that we may get here is that the `composableCow` contract actually isn't a composableCow contract containing the `getTradeableOrderWithSignature` function. This will lead to `EvmError` issues.

One way to get around this would be in the `addContract` to actually filter on the basis that we *expect* the `getTradeableorderWithSignature` function to exist, so check for the function's selector in `getCode` for the address.

This is a suggestion to help reduce potential issues :smile:

mfw78 (approved):
Approve with a suggestion of how one can filter the contracts to remove those that don't comply (ie. don't implement `getTradeableOrderWithSignature`).

actions/checkForAndPlaceOrder.ts
anxolin: It's the linter. It is the same as not having it :) 
i think is prettier in action. I actually also found it weird, but don't argue to much with them.

You can see it in action here: https://codebeautify.org/typescript-formatter-online



image

actions/checkForAndPlaceOrder.ts
anxolin: 100%, we should do this, it is a problem that we show too many errors. 
Ideally this project should only throw when there's sth meaningful to complain about. 

I'll add the issue and probably reiterate soon on that soon.

anxolin: https://github.com/cowprotocol/composable-cow/issues/50

anxolin:
Addressed the comments and merging. 

Pls let me know if you need to re-iterate again, i will follow up in another PR.

PR #47, opened 2023-08-18
Raw broadcast folder

fleupold:
Still needs to be pruned to only contain the relevant deployments


mfw78 (approved):

mfw78:
This PR has been cleaned so that:

1. The `out` now contains just the source ABI files related specifically to `ComposableCoW`.
2. The `broadcast` directory now contains the deployment locations, and receipts.

mfw78 (approved):

PR #56, opened 2023-08-20
Validate orders using custom handlers

anxolin:
This PR adds some basic mechanisms to introduce custom validations for specific handlers

The idea is to be able to verify things like: "TWAP order is already expired' which will allow us to not check for this order again 

This PR just adds the placeholder to introduce the logic, and defines the interfaces of the validation.

I will leave some comments in the code to explain this PR. 

## checkForAndPlaceOrders
Now this action will use a new method called `validateOrder`.

It will allows this action to perform some specific smart order validations without introducing specific logic in the main actions logic. 

All this algorithm will care is about the result of the validation. The specific implementation will signal if the order should be deleted (so its not checked again), and if there was an error doing the validation, and if this error was expected or unexpected.

Expected errors are for example: is not yet time to place an order

Unexpected errors are for example: Error querying the node

Expected errors will not result in Tenderly to fail. It will still show success. Also we won't try to do the CALL to `getTradeableOrderWithSignature` so this will reduce a lot the lofa

## cow-SDK
This PR adds a dependency with the SDK, so it can make use of its logic. 

Ideally, the validation logic should live there. Also, future PRs should consider using the OrderBook API through the SDK and not though an ad-hoc axios request.

## For more details
Check the code, i will add some comments there

### DIFF
--- a/actions/checkForAndPlaceOrder.ts
+++ b/actions/checkForAndPlaceOrder.ts
@@ -18,8 +18,15 @@ import {
   init,
   writeRegistry,
 } from "./utils";
-import { ChainContext, ConditionalOrder, OrderStatus } from "./model";
+import {
+  ChainContext,
+  ConditionalOrder,
+  OrderStatus,
+  SmartOrderValidationResult,
+  ValidationResult,
+} from "./model";
 import { GPv2Order } from "./types/ComposableCoW";
+import { validateOrder } from "./handlers";
 
 const GPV2SETTLEMENT = "0x9008D19f58AAbD9eD0D60971565AA8510560ab41";
 
@@ -148,6 +155,24 @@ async function _processConditionalOrder(
       }
     }
 
+    // Do custom Conditional Order checks
+    const [handler, salt, staticInput] = await (() => {
+      const { handler, salt, staticInput } = conditionalOrder.params;
+      return Promise.all([handler, salt, staticInput]);
+    })();
+    const validateResult = await validateOrder({
+      handler,
+      salt,
+      staticInput,
+    });
+    if (validateResult && validateResult.result !== ValidationResult.Success) {
+      const { result, deleteConditionalOrder } = validateResult;
+      return {
+        error: result !== ValidationResult.FailedButIsExpected, // If we expected the call to fail, then we don't consider it an error
+        deleteConditionalOrder,
+      };
+    }
+
     // Get GPv2 Order
     const tradeableOrderResult = await _getTradeableOrderWithSignature(
       owner,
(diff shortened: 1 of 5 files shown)


actions/checkForAndPlaceOrder.ts
```diff
@@ -148,6 +155,24 @@ async function _processConditionalOrder(
       }
     }
 
+    // Do custom Conditional Order checks
+    const [handler, salt, staticInput] = await (() => {
+      const { handler, salt, staticInput } = conditionalOrder.params;
+      return Promise.all([handler, salt, staticInput]);
+    })();
+    const validateResult = await validateOrder({
```
anxolin, line 163: We validate the static input, if the handler is know. 

This action doesn't care about the specific validation, only the result of it

actions/handlers/index.ts
```diff
@@ -0,0 +1,27 @@
...
+const validatorsForHandler: Record<string, SmartOrderValidator> = {
+  [handlerTwap.toLowerCase()]: validateOrderTwap,
+};
+
+/**
+ * Validate a smart order using the custom handler logic (if known)
+ *
+ * @returns The result of the validation if the handler is known, undefined otherwise
+ */
+export async function validateOrder(
```
anxolin, line 21: generic function, it will delegate to the right validator

actions/handlers/index.ts
```diff
@@ -0,0 +1,27 @@
...
+  ValidateOrderParams,
+} from "../model";
+
+import {
+  handlerAddress as handlerTwap,
+  validateOrder as validateOrderTwap,
+} from "./twap";
+
+const validatorsForHandler: Record<string, SmartOrderValidator> = {
+  [handlerTwap.toLowerCase()]: validateOrderTwap,
```
anxolin, line 13: we register one validator per known handler

actions/handlers/twap.ts
```diff
@@ -0,0 +1,14 @@
+import { TWAP_ADDRESS } from "@cowprotocol/cow-sdk";
+
+import { SmartOrderValidator, ValidationResult } from "../model";
+
+export const handlerAddress = TWAP_ADDRESS;
+
+export const validateOrder: SmartOrderValidator = async (_params) => {
+  // TODO: Implement validation logic, ideally delegate to the SDK
```
anxolin, line 8: This PR don't implement the specic logic, this could come in future PRs. 

As hinted here, most of the logic would ideally come from the SDK. This validator would just transform it in the result that this function expects

actions/package.json
```diff
@@ -18,6 +18,7 @@
   },
   "dependencies": {
     "@cowprotocol/contracts": "^1.4.0",
+    "@cowprotocol/cow-sdk": "^2.4.0-rc.0",
```
anxolin, line 21: Added the SDK, as @mfw78 pointed out, it should be ideally the source of truth for JS logic

mfw78 (approved):

PR #58, opened 2023-08-21
fix: flaky test

mfw78:
This PR:

1. Applies a small fix to enforce `SafeCast` conditions on the fuzz test's parameters.


mfw78:
This was merged in with the abi artifacts / deployment info.

PR #61, opened 2023-09-02
fix: standardise openzeppelin import path

mfw78:
**Problem**

When developing integrations, most libraries that import openzeppelin contracts do so using `npm` scope (ie. `@openzeppelin/contracts`). Using non-standard import paths can create complexities with dependency management / remappings.

**Solution**

For `openzeppelin`, standardise around the use of `@openzeppelin/contracts` so those building on top of these repositories don't have the aforementioned problems.


fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2FpWv5zAaHdX5QY%2Fgiphy.gif&token=housekeeping" data-gitmeme-token="housekeeping"></a>

PR #63, opened 2023-09-02
Update GoodAfterTime.sol:Improved Struct Packing

Aryan9592:
This article will be helpful to understand the reason behind the PR -> https://dev.to/javier123454321/solidity-gas-optimizations-pt-3-packing-structs-23f4

For more detailed understanding refer this -> https://github.com/Consensys/linea-attestation-registry/pull/101

### DIFF
--- a/src/types/GoodAfterTime.sol
+++ b/src/types/GoodAfterTime.sol
@@ -36,11 +36,11 @@ contract GoodAfterTime is BaseConditionalOrder {
         IERC20 sellToken;
         IERC20 buyToken;
         address receiver;
+        bool allowPartialFill;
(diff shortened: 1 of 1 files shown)


mfw78:
Hey, thanks for the PR on the gas optimisation. From my understanding, this isn't required as the methodology for tightly packing structs is useful when the structs are stored using `SSTORE`, however these data structs in order types are only read from `calldata` / written back and forth in memory, so there's no `SSTORE`-related packing benefits.

Aryan9592:
Well, in that case this PR ain't necessary to be merged. Thanks for pointing this out!

PR #66, opened 2023-09-17
fix: cicd secret

mfw78:
This PR:

1. Fixes the CI/CD secret to one that has the correct scope.


mfw78:
As this is narrow in scope, being only related to project management, will merge without reviews.

PR #73, opened 2023-09-18
feat: watch tower / sdk friendly expanded interface

### LINKED ISSUE #72: feat: watch-tower and sdk friendly extended interface
### Problem
It's frustrating when wanting to implement a watch tower for conditional orders as the current interfaces do not provide any method of communicating their intent to the watch tower (such as try again at another block, another epoch etc).

### Suggested solution
Revise the interface of custom errors to include parameterized custom errors capable of communicating target epochs, block numbers etc at which the watch tower should attempt to get a discrete order again - or even delete the conditional order from monitoring!

### Alternatives considered
N/A

### Additional context
This is the result of much iteration on the Tenderly Web3 actions watch-tower and limitations that have been observed in production.

### Acceptance criteria

- [x] Interface for "try at epoch"
- [x] Interface for "try at block number"
- [x] Communicate error messages (strings) from smart contracts accurately to the watch tower

mfw78:
# Description

Provides expanded `IConditionalOrder` interfaces with parameterized errors as these are required to standardise such that the conditional order is able to communicate it's indexing intent to a watch tower.

# Changes

- [x] Add custom parameterized errors for watch tower indexing.
- [x] `validateData` external function for verifying a conditional order's data struct if granular verification has been implemented (recommended for order type developers).

## How to test
The TWAP contract has been updated to implement the expanded interface. To test:

```bash
forge test
```

## Related Issues

Fixes #72


src/types/twap/TWAP.sol
```diff
@@ -59,4 +59,12 @@ contract TWAP is BaseConditionalOrder {
             revert IConditionalOrder.OrderNotValid(NOT_WITHIN_SPAN);
         }
     }
+
+    /**
+     * @inheritdoc IConditionalOrder
+     */
+    function validateData(bytes memory data) external override pure {
+        TWAPOrder.validate(abi.decode(data, (TWAPOrder.Data)));
```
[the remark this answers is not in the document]
mfw78: The intent was to not change existing contracts, however, as the interface enforces this, I'll add the changes to TWAP, but remove the deployment addresses in `README.md` as the changes are not reflective of what is deployed (and we need to be careful and ensure we don't merge these into `main` unless there is a clear intent to have the changes audited).

tl;dr I don't think it's reasonable to count on these changes being usable in production at the moment until such time in the future that there are additional order types / tweaks that necessitate a new audit / review.

src/interfaces/IConditionalOrder.sol
```diff
@@ -14,6 +14,16 @@ interface IConditionalOrder {
     ///      A parameter of `string` type is included to allow the caller to specify the reason for the failure.
     error OrderNotValid(string);
 
+    // --- errors specific for polling
+    // Signal to a watch tower that polling should be attempted again.
+    error PollTryNextBlock(string reason);
```
fleupold, line 19: I wonder if this information should/could already be emitted when the Conditional Order is created?

fleupold (commented):
Nice addition. Any thoughts on how we can enforce/incentivise developers to care about choosing the most suitable hint (if I don't care about watchtower costs, I'd let it poll every block).

src/interfaces/IConditionalOrder.sol
mfw78: Great idea, and certainly valid for single order types. I've added the issue #75 to track this initiative. Worst case scenario in the current scheme, we consume one additional poll cycle on each conditional order creation. I think it would be reasonable for us to defer this to avoid widening the scope too much for the valory watch-tower adaption (would necessitate monitoring two events, given the TWAP legacy).

mfw78:
> Nice addition. Any thoughts on how we can enforce/incentivise developers to care about choosing the most suitable hint (if I don't care about watchtower costs, I'd let it poll every block).

Incentives research is tracked [here](#76).

PR #78, opened 2023-10-01
tests: revert with more than 4bytes

mfw78:
# Description
The current test suite only allows for testing settlements that may revert with only a `4bytes` selector (ie. no parameterized). This PR remedies that.

# Changes

- [x] Abstract revert data can be asserted with `settle` in the test harness.
- [x] Standardises the use of `hex""` for hex literal strings / empty bytes.
- [x] Gardening to dry some test of the test harness.

## How to test

1. `forge test -vvv --optimizer-runs 200 --no-match-test "fork|simulate" --fuzz-seed 672679878`

### DIFF
--- a/src/types/twap/TWAP.sol
+++ b/src/types/twap/TWAP.sol
@@ -76,8 +76,7 @@ contract TWAP is BaseConditionalOrder {
     /**
      * @inheritdoc IConditionalOrder
      */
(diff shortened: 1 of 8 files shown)


mfw78:
As this PR is only related to the test harness with a narrow scope on increasing the usability of `settle`, with all tests in CI/CD passing, will merge to `develop` without review.

PR #83, opened 2024-01-10
Add Sepolia deployments

fedgiac:
# Description

This PR records the deployment of the composable cow contracts on Sepolia.

To guarantee that the contracts deploy to the same deterministic addresses work, we needed to restore the remapping to those used in earlier deployments: remappings are included in the Solidity metadata, which change the bytecode and in turn the addresses.

# Changes

- Sepolia deployment
- Update readme with new contracts and information on how to deploy

## How to test

Check verified contract code on Etherscan and compare the addresses in the readme with those in the `run-latest.json`.


anxolin:
I believe @fleupold might know how to get the same address as he managed for the other networks

Anyways, although desirable, I'm fine with a different address.

fleupold:
I don't remember what I did, but looking at the two simulated create transactions
1. [Sepolia](https://dashboard.tenderly.co/tx/sepolia/0xe344f7fab6e56d6ca4aa8927ba04b74c0074e29878a2da84813f91015ca1c27c?trace=0)
2. [Mainnet](https://dashboard.tenderly.co/tx/mainnet/0x2d4159ae0d4f026e2063292f3e40d898982619853484b8cd84f3748503fcdd19?trace=0)

you can see that the calldata differs slightly (the 64bytes starting from index 17484) which is likely causing the different addresses. 

```
0a02524bcc29f556711d99bb4549a8aaa69e26b88360aa4b69697bff89a56305
0e23d95217a10665974c83c4ea22bd347daab0570e397236749f8f5063e1b702
```

In the worst case we could manually deploy using the exact same calldata, but maybe we can get to the bottom of this (looks like it's the last bytes of the compiled bytecode that is different)

mfw78:
OK, we went on a deep dive here and there were some learnings.
1. The differences in the `bytecode` observed are related to a metadata hash between the `initcode` and the constructor arguments.
2. `forge` passes in the `remappings` dict that is used in the course of compiling as part of the `metadata` which then makes it into the hash, affecting deterministic deployments.
3. #47 was analysed for where the deployment data was committed to the repository. By then narrowing down to 559780acf9d6c0f39812366e4b53bfbb0729266b, it was possible to obtain the remappings dict that was used from the `metadata` output of `ExtensibleFallbackHandler.json` (the assumption is that it was the same at the point of running the deploy script for all, which it was).
4. Recreate the `remappings.txt`, compile and, voila same contract address on Sepolia.

Learnings: For some reason, the `remappings.txt` has diverged / wasn't correct at the time of the `v1.0.0` tag commitment. Attention needs to be paid to this aspect when dealing with subsequent deployments.

mfw78 (approved):
Glad we got this sorted!

mfw78 (approved):

PR #87, opened 2024-05-07
Refactor handler errors to try on next block watch tower pattern

### LINKED ISSUE #85: chore: Interface with errors supported by the watchtower
### Background
The watchtower has a wide range of custom errors that can be used to tell it if and when an order should be recreated (see [here](https://github.com/cowprotocol/watch-tower/blob/b9bf2a9da33dd51e99c987ba3c8f39d22588c536/src/utils/contracts.ts#L100-L110)). They are also documented [here](https://docs.cow.fi/cow-protocol/reference/contracts/periphery/composable-cow#custom-error-codes).
The only error that's available on an interface is `OrderNotValid`.

### Details
It's helpful for developers to be able to refer to these errors without having to search for them and include them in their handler code.

### Acceptance criteria

Either of:
- [x] new interface with all custom errors accepted by the watchtower, or
- [ ]  include the errors in `IConditionalOrderGenerator`.

yvesfracari:
# Description
Refactor Stop Loss, Trade Above Threshold, and Good After Time errors to watch tower try again pattern.

# Changes
- [x] Add `revertPollAtNextBlock`method on `ConditionalOrdersUtilsLib`
- [x] Stop Loss Refactoring
- [x] Trade Above Threshold Refactoring
- [x] Good After Time Refactoring

## How to test
1. Run modified test files
2. Stop Loss Deployment on sepolia: `0x410c267baf50b36475B30051Da2aE734f6726F01`

## Related Issues
- Fix #85


yvesfracari:

mfw78:
In general, I'd suggest to include all the watch-tower errors that are possible. I'd already done some work in `develop` on this. Refer: https://github.com/cowprotocol/composable-cow/blob/24ce6fb9350755bdc71f3e02bc7df7b781b73fcb/src/interfaces/IConditionalOrder.sol#L17-L25

Instead of introducing this within `IConditionalOrder` as I did, having a separate interface, as you have done, for the watch-tower custom errors is fine.

src/types/ConditionalOrdersUtilsLib.sol
```diff
@@ -30,4 +32,15 @@ library ConditionalOrdersUtilsLib {
...
+    /**
+     * @dev Reverts call execution with a custom error that indicates to the
+     * watchtower to poll for new order when the next block is mined.
+     */
+    function revertPollAtNextBlock(string memory message) internal view {
+        revert IWatchtowerCustomErrors.PollTryAtBlock(
+            block.number + 1,
+            message
+        );
+    }
```
mfw78, line 45: I'd generally discourage the use of revert utilities, such as this, to reduce the layers of indirection required when reviewing the logic within the smart contract to know where I can expect a revert (sure, it's in the name, but it's another function that I then need to drill into to triple check that it indeed does revert). Depending on the optimiser as well and how this gets compiled, this may also result in additional gas consumption.

mfw78 (commented):
In general looks OK. I've left some comments about `IwatchtowerCustomErrors` in the main PR comments. FYI, there is a `error PollTryNextBlock(string reason)` that is supported by the watch-tower. This allows for simpler code as it was identified that trying at the very next block is a common pattern and alleviates a lot of `block.number + 1`.

mfw78 (approved):
LGTM

ribeirojose:
rebased onto https://github.com/cowprotocol/composable-cow/pull/88 @mfw78

yvesfracari:
@mfw78 since we're planning to launch the stop-loss app soon, I think that we should have this PR merged. Let me know if you need more things on our side. All the changed handlers contracts will need to be redeployed, I was assuming that you guys will do it since you have a wallet to do this process, right?

mfw78 (dismissed):

src/types/GoodAfterTime.sol
```diff
@@ -67,7 +67,7 @@ contract GoodAfterTime is BaseConditionalOrder {
 
         // Require that the sell token balance is above the minimum.
         if (!(data.sellToken.balanceOf(owner) >= data.minSellBalance)) {
-            revert IConditionalOrder.OrderNotValid(BALANCE_INSUFFICIENT);
+            revert IConditionalOrder.PollTryNextBlock(BALANCE_INSUFFICIENT);
```
fedgiac, line 70: Wasn't the old behavior intended? In this contract's `@dev` comments I read:

> This order type allows for placing an order [...]. As [...], this allows for placing multiple orders. To ensure that the order is not filled multiple times, a `minSellBalance` is checked before the order is placed.

My understanding is that trying again should be avoided because, once the order is matched, the balance would go below `minSellBalance` and (I suppose) should never go back up.

fedgiac (commented):
The changes look good to me with the exception of the comment below. Once clarified it's good for merging.

src/types/GoodAfterTime.sol
mfw78: Great catch! ️🕵️

src/types/GoodAfterTime.sol
```diff
@@ -62,7 +62,7 @@ contract GoodAfterTime is BaseConditionalOrder {
 
         // Don't allow the order to be placed before it becomes valid.
         if (!(block.timestamp >= data.startTime)) {
-            revert IConditionalOrder.OrderNotValid(TOO_EARLY);
+            revert IConditionalOrder.PollTryNextBlock(TOO_EARLY);
```
fedgiac, line 65: This looks like a nice use-case for `PollTryAtEpoch`:

```suggestion
            revert IConditionalOrder.PollTryAtEpoch(data.startTime, TOO_EARLY);
```

fedgiac (approved):
Nice!

PR #89, opened 2024-07-24
Refactor stop-loss create just 1 discrete order

yvesfracari:
# Description
- Avoid multiple discrete order creations for one stop-loss order creation.

# Changes
- [x] Replace the `validityBucketTime` parameter of stop loss data with constant `validTo`.

## How to test
1. Run the modified local test.
2. Try to create multiple discrete orders from the same stop-loss order using the deployed contract on sepolia: `0xe6CDbC068654C506424F7747357F51d0e7caB00e`

### DIFF
--- a/src/types/StopLoss.sol
+++ b/src/types/StopLoss.sol
@@ -13,6 +13,8 @@ string constant ORACLE_INVALID_PRICE = "oracle invalid price";
 string constant ORACLE_STALE_PRICE = "oracle stale price";
 /// @dev The strike price has not been reached
 string constant STRIKE_NOT_REACHED = "strike not reached";
+/// @dev The order is not valid anymore
+string constant ORDER_EXPIRED = "order expired";
 
 /**
  * @title StopLoss conditional order
@@ -34,7 +36,7 @@ contract StopLoss is BaseConditionalOrder {
      * @param receiver: The account that should receive the proceeds of the trade
      * @param isSellOrder: Whether this is a sell or buy order
      * @param isPartiallyFillable: Whether solvers are allowed to only fill a fraction of the order (useful if exact sell or buy amount isn't know at time of placement)
-     * @param validityBucketSeconds: How long the order will be valid. E.g. if the validityBucket is set to 15 minutes and the order is placed at 00:08, it will be valid until 00:15
+     * @param validTo: The UNIX timestamp that sets how long the order will be valid for
      * @param sellTokenPriceOracle: A chainlink-like oracle returning the current sell token price in a given numeraire
      * @param buyTokenPriceOracle: A chainlink-like oracle returning the current buy token price in the same numeraire
      * @param strike: The exchange rate (denominated in sellToken/buyToken) which triggers the StopLoss order if the oracle price falls below. Specified in base / quote with 18 decimals.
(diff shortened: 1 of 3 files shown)


src/types/StopLoss.sol
```diff
@@ -34,7 +36,7 @@ contract StopLoss is BaseConditionalOrder {
      * @param receiver: The account that should receive the proceeds of the trade
      * @param isSellOrder: Whether this is a sell or buy order
      * @param isPartiallyFillable: Whether solvers are allowed to only fill a fraction of the order (useful if exact sell or buy amount isn't know at time of placement)
-     * @param validityBucketSeconds: How long the order will be valid. E.g. if the validityBucket is set to 15 minutes and the order is placed at 00:08, it will be valid until 00:15
+     * @param validTo: The UNIX timestamp that sets how long the order will be valid for
```
ribeirojose, line 39:
```suggestion
     * @param validTo: The UNIX timestamp until when this order is valid
```

test/ComposableCoW.stoploss.t.sol
```diff
@@ -269,6 +272,46 @@ contract ComposableCoWStopLossTest is BaseComposableCoWTest {
...
+            buyAmount: 1 ether,
+            appData: APP_DATA,
+            receiver: address(0x0),
+            isSellOrder: false,
+            isPartiallyFillable: false,
+            validTo: validTo,
+            maxTimeSinceLastOracleUpdate: 15 minutes
+        });
+
+        vm.expectRevert(
```
fedgiac, line 300: It would be better if there was also a test that shows an order is valid if warping to `validTo`, all other tests use `type(uint256).max` for the valid to so the positive boundary condition isn't tested on a different number.

fedgiac, line 300: It would be better if there was also a test that shows an order is valid if warping to `validTo`, all other tests use `type(uint256).max` for the valid to so the positive boundary condition isn't tested on a different number.

fedgiac (commented):
The changes look good, but why was it previously implemented in bucket?
My only guess is that maybe long-lived orders aren't accepted by the API. In case, has this changed?

mfw78:
> The changes look good, but why was it previously implemented in bucket? My only guess is that maybe long-lived orders aren't accepted by the API. In case, has this changed?

Because I was foolish when I implemented it :see_no_evil:

yvesfracari:
@mfw78 once this is merged, let us know if we should deploy this contract on all chains or you will do it.

test/ComposableCoW.stoploss.t.sol
```diff
@@ -284,15 +287,61 @@ contract ComposableCoWStopLossTest is BaseComposableCoWTest {
         stopLoss.getTradeableOrder(safe, address(0), bytes32(0), abi.encode(data), bytes(""));
     }
 
-    function test_strikePriceMet_fuzz(int256 sellTokenOraclePrice, int256 buyTokenOraclePrice, int256 strike) public {
+    function test_OracleRevertOnExpiriedOrder_fuzz(
```
fedgiac, line 290:
```suggestion
    function test_OracleRevertOnExpiredOrder_fuzz(
```

fedgiac (approved):

mfw78 (approved):

fedgiac (approved):

PR #91, opened 2024-08-16
New handler deployments

mfw78:
# Description

Redeploys handlers to correctly respect the watch-tower interface.

# Changes

- [x] Renamed previous deployments (broadcast)
- [x] Deployed new handlers from recent changes #87 and #89 
- [x] Updated README
- [x] Added bash script for generating `networks.json`
- [x] Generated `networks.json`
- [x] Removed Goerli links due network being deprecated
- [x] Added Gnosis supplementary audit

## How to test

1. Verify that respective contracts have been redeployed and are verified.


dev/generate-network-files.sh
```diff
@@ -0,0 +1,24 @@
...
+    | select(.transactionType == "CREATE2")
+    | select(.hash != null)
+    | {(.contractName): {($chainId): {address: .contractAddress, transactionHash: .hash }}}
+  '  <"$deployment"
+done \
+  | # Then, all these single-contract single-chain-id networks.jsons are merged. Note: in case the same contract is
+    # deployed twice in the same script run, the last deployed contract takes priority.
+    jq --sort-keys --null-input 'reduce inputs as $item ({}; . *= $item)'
+
+# Todo: handle case where the same contract is deployed on multiple chains.
```
fedgiac, line 24: I think this isn't needed anymore?

```suggestion
```

dev/generate-network-files.sh
```diff
@@ -0,0 +1,24 @@
...
+  # First, every single deployment is formatted as if it had its own networks.json
+  jq --arg chainId "$chain_id" '
+    .transactions[]
+    | select(.transactionType == "CREATE2")
+    | select(.hash != null)
+    | {(.contractName): {($chainId): {address: .contractAddress, transactionHash: .hash }}}
+  '  <"$deployment"
+done \
+  | # Then, all these single-contract single-chain-id networks.jsons are merged. Note: in case the same contract is
+    # deployed twice in the same script run, the last deployed contract takes priority.
```
fedgiac, line 21: Comment to make the implicit assumption in the loop explicit.

```suggestion
  | # Then, all these single-contract single-chain-id networks.jsons are merged. Note: in case the same contract is
    # deployed twice in the same script run, the last deployed contract takes priority.
    # If the same contract is deployed twice in different runs, the address in the file path that comes latest in
    # alphabetical order takes priority. For example, a contract in `broadcast/Deployment10/*` is overwritten by
    # one with the same name from `broadcast/Deployment2/*`.
```

fedgiac (commented):
In general it looks good with a small caveat (see not-comment-related comment :slightly_smiling_face:).

The bytecode when building on my local machine matches what I see onchain for all addresses.
Readme update is consistent with the new addresses.
`networks.json` is consistent with the output of the new script (with caveat).

mfw78:
As this is related only to deployment addresses, merging with 1 review.

PR #99, opened 2025-03-24
Add draft for custom twap order type with increasing minPartLimit for each part

najienka:
# Description
Based on a [feedback](https://x.com/tatsugitsune/status/1897399822266851764) received for limit orders. The feedback is as follows:
```Create a feature that allows a user to generate multiple limit orders with one signed statement, where each limit order buys or sells the same number of tokens in 1% increments. 

For example, create 50 sell limit orders from price A up to price B, with X number of tokens, and ensure that each iteration is 1% higher than the previous limit order. Divide the tokens equally across a price range.

This will allow users to efficiently dollar cost average in and out of a market without having to sign 50 messages.
```

## Changes

This PR introduces a draft smart contract, `TWAPWithIncreasingMinPartLimit.sol`, located in `src/types`.

The contract implements a custom TWAP (Time-Weighted Average Price) limit order where:

1. The percentage increment and minimum buy amount for each part are dynamically calculated.

2. These calculations use the specified start and end prices of the TWAP.

3. A unique order identifier determines the specific TWAP part for which a conditional order is generated.

4. This enables automated, structured execution of TWAP orders with increasing limits over time.

### DIFF
--- a/src/types/BatchLimitSell.sol
+++ b/src/types/BatchLimitSell.sol
@@ -27,7 +27,7 @@ string constant INVALID_TOKEN = "invalid token";
 string constant INVALID_PART_SELL_AMOUNT = "invalid part sell amount";
 /// @dev The minimum buy amount in each part (limit) is not gretater than zero.
 string constant INVALID_MIN_PART_LIMIT = "invalid min part limit";
-/// @dev The valid until time is 
+/// @dev The valid until time is invalid.
 string constant INVALID_VALID_TO_TIME = "invalid valid until time";
 
 /// @title BatchLimitSell Conditional Order
@@ -76,7 +76,7 @@ contract BatchLimitSell is BaseConditionalOrder {
 
         _validate(data); 
 
-        // Get the sell price for the part
+        // Get the min part limit for the part
         uint256 minPartLimit = _getMinPartLimit(data);
 
         if (!(minPartLimit > 0)) revert IConditionalOrder.OrderNotValid(INVALID_MIN_PART_LIMIT);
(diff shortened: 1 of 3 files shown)


najienka:

src/types/TWAPWithIncreasingMinPartLimit.sol
```diff
@@ -0,0 +1,147 @@
...
+
+    struct Data {
+        IERC20 sellToken; // The token being sold
+        IERC20 buyToken; // The token being bought
+        address receiver; // The order owner (user who initiated the order)
+        uint256 partSellAmount; // The amount to sell in each part
+        uint256 startPrice; // Sell price for the first part
+        uint256 endPrice; // Sell price for the last part
+        uint256 t0; // Start timestamp
+        uint256 t; // Time interval for each part
```
fleupold, line 37: Reading the suggested order type (from the tweet you linked), I don't really understand how time is coming into play here? I understand the user wants to place n orders with one signature to buy at different price points (e.g. buy 1% at $1, 1% at $1.01, 1% at $1.02 ... $2).

If the price moves, all orders in the price range should be executed (there shouldn't be any notion of start time)

fedgiac (commented):
It would be nice to see this order type actually being executed (with, say, `n = 2` for simplicity) on a test transaction.

src/types/TWAPWithIncreasingMinPartLimit.sol
najienka: Remove `startTime` and `endPrice` and updated the logic to get the percentage increase from off-chain and compute the `minPartLimit ` for each part.

src/types/BatchLimitSell.sol
```diff
@@ -0,0 +1,132 @@
...
+    {
+        // Decode the payload into the twap with increasing minPartLimit parameters.
+        Data memory data = abi.decode(staticInput, (Data));
+
+        /**
+         * @dev If `data.validTo` is set to 0, then get the validTo time from the context.
+         */
+        if (data.validTo == 0) {
+            data.validTo = uint256(COMPOSABLE_COW.cabinet(owner, ctx));
+        }
```
fleupold, line 75: Can you explain why this is needed for this order type?

fedgiac:
Closing this PR as development of this feature has stopped.

PR #116, opened 2026-06-30
feat: Just in time polling of funds (PoC)

anxolin:
> 🧩 This PoC is being broken up into reviewable stacked PRs — tracked in #121 (see #122, #123, #124, #125).

## What

Adds `ComposableCowPoller`, a small contract that funds a conditional order **just in time**, right before each part settles, instead of locking up the full amount when the order is created.

I don't expect this PR to be merged as it is, my goal is to create a PoC and test the results of the exploration done on ways to implement JIT polling of funds (see  [private notes](https://app.notion.com/p/cownation/TWAP-for-EOA-private-notes-36c8da5f04ca80798a4fed91cda9bfcc?source=copy_link)).

This is a follow up on https://github.com/cowdao-grants/cow-shed/pull/53 and https://github.com/anxolin/cow-sdk-scripts/pull/12 ([slack](https://nomevlabs.slack.com/archives/C036G0J90BU/p1757584019667809))

## Why

For order types that run over long periods (e.g. TWAP, specially when using it for DCA), pre-funding the entire thing up front is capital-inefficient and sometimes impossible: the funds may only arrive shortly before each part needs to trade.

This contract lets capital stay in the user's own account (an EOA or treasury) and pulls only what the current part needs, only when it needs it.

## Show me the proof
If you want to check this end-to-end working, I placed an example order here:

- **sell=buy (create TWAP)** https://explorer.cow.fi/gc/orders/0xcc1382bce5fb733e9cb235f6536a152d3e2acc663ac5e5f7a4b33dbc0316f6b4016f34d4f2578c3e9dffc3f2b811ba30c0c9e7f36a44505d?tab=overview
- **TWAP 1/2:**: Pull and sell 0.1 sDAI for at least 0.77 COW, recipient EOA, trader cow-shed
https://explorer.cow.fi/gc/orders/0x24106d3b23805bd9b992bfdba43bd692de4582180acf44ca4edb0d698312562d62587918b2f00176646679509217a5a4d1ebbfd56a444b69
- **TWAP 2/2:**: Pull and sell 0.1 sDAI for at least 0.77 COW, recipient EOA, trader cow-shed https://explorer.cow.fi/gc/orders/0x693a30318b8208f78b5028b237bb3cf3cef8c7c7c5a74773b02d4a0dec6f334762587918b2f00176646679509217a5a4d1ebbfd56a444b69

## How it works

A funder registers a *schedule* for a given order, pointing at:
- the order's handler (e.g. the TWAP type),
- the funds source (`funder`),
- the order owner that will hold and sell the funds,
- the order's `salt`, and
- the order's `staticInput`.

The schedule is stored under an appData-independent key `id = scheduleId(funder, handler, owner, salt)`.

Anyone can then call `topUp(id)` (designed to run as a CoW pre-hook). The contract re-derives the order's `ctx` on-chain from the stored params and pulls exactly the current part's `sellAmount` from the funder into the owner. It is deliberately generic: it works for any conditional order, with no order-type-specific logic of its own.

### Design note: making `topUp` embeddable in the order's own appData

The natural identifier for a schedule is the order's `ctx` (`ComposableCoW.hash(params)`). But the goal is to trigger `topUp` from a **pre-hook in the order's own appData**, so each discrete part funds itself just-in-time and that creates a cycle:
* `ctx` contains the order's `appData` hash → but the hook (which references the schedule) lives *inside* that `appData`. 
* Any order-derived identifier (`ctx`, `staticInput`, or the order digest) has the same problem, since they all transitively contain `appData`. 

The fix is to key the schedule by data that is **independent of `appData`**: `funder`, `handler`, `owner`, `salt`

The poller should be able to reconstruct the `ctx` on-chain inside `topUp` if it has the `handler`, `salt` and `staticinput` (this is why is stored in the schedule)

The hook calldata `topUp(id)` then no longer depends on `appData`, breaking the cycle. Every discrete part inherits the parent order's `appData`, so each part self-funds via its own pre-hook.

`topUp` is permissionless (anyone can call it, but meant to be called by the trampoline). 

Safety comes from constraining *what* the call can do, not *who* makes it:

1. The order must still be authorised in `ComposableCoW` (cancelling it disables the poller). `ctx` is re-derived on-chain from the stored params, never taken from the caller.
2. The amount, sell token, and validity window are read from the handler's own `getTradeableOrder` (never from the caller). Outside an active window that call reverts, so nothing can be pulled when no part is tradeable.
3. Funds can **only ever move to the owner registered in the schedule**.
4. Each part is funded at most once, keyed by its unique discrete order digest. This prevents a future part's funds from being pulled early. For TWAP was enough to check the deadlline, but this way is generic for more orders!
5. The top-up is balance-capped to one part, so repeated calls are idempotent.
6. The schedule `id` is namespaced by `funder`, so nobody can register, overwrite, or squat another funder's schedule.

Only the funder can register or revoke a schedule, so nobody can point a pull at someone else's funds.

## How to review
- Review `src/types/ComposableCowPoller.sol` which is the main contract. You might want to review my [private notes](https://app.notion.com/p/cownation/TWAP-for-EOA-private-notes-36c8da5f04ca80798a4fed91cda9bfcc?source=copy_link) on the exploration I did for the options to arrive to this proposal.
- A contract was deployed at https://gnosisscan.io/address/0xA360eE11eD0d2025604518CF4B8F6e6CB76C7Df7.
- Rest of the PR is some deployment scripts and unit tests, which cover:
  - the happy path (each part funded just in time across the whole schedule)
  - idempotency within a part
  - the guard against funding a future part early
  - reverting outside the schedule window or with no schedule
  - disabling via order cancellation
  - the funder-only register/revoke checks.

## Out of scope
I believe we can do some enhancements to use signatures for registering the schedule. This would allow us to register during the sell=buy order execution, as part of creating the TWAP! 

## Test plan

```bash
```


anxolin:

anxolin:
Closing, as this was just a PoC. Created an issue for implementing it https://github.com/cowprotocol/composable-cow/issues/121

Issue #121, opened 2026-07-10
Just in time polling of funds


anxolin:
`ComposableCowPoller` funds a conditional order just in time: it pulls exactly the current part's `sellAmount` from a funds source into the order owner right before that part settles, so capital isn't locked up front for the whole order (e.g. a long TWAP or DCA). See #116 for the design and rationale.

#116 landed it all in one PoC PR. To make it reviewable, the work is split into small PRs, each with only its own tests:

- [ ] #122 empty `ComposableCowPoller` contract and test
- [ ] #123 register a polling schedule
- [ ] #124 revoke a schedule
- [ ] #125 poll the order's funds just in time
- [ ] #126 deploy the contract

PR #122, opened 2026-07-10
feat: base ComposableCowPoller contract

anxolin:
Part of #121.

## What
Introduces the base `ComposableCowPoller` contract as an empty shell (no state, no functions yet) plus the empty `test/ComposableCowPoller.t.sol` scaffolding. 

This is the base for follow up PRs which will add features to the contract.

## Test plan
```bash
```
Test should pass

### DIFF
--- a/test/ComposableCowPoller.t.sol
+++ b/test/ComposableCowPoller.t.sol
@@ -18,6 +18,6 @@ contract ComposableCowPollerTest is BaseComposableCoWTest {
     }
 
     function test_deployment() public {
-        assertTrue(address(poller) != address(0), "poller deployed");
+        assertTrue(address(poller).code.length > 0, "poller deployed");
     }
 }
--- a/test/ComposableCowPoller.t.sol
+++ b/test/ComposableCowPoller.t.sol
@@ -1,0 +1,23 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity >=0.8.0 <0.9.0;
+
+import {BaseComposableCoWTest} from "./ComposableCoW.base.t.sol";
+
+import {ComposableCowPoller} from "../src/types/ComposableCowPoller.sol";
+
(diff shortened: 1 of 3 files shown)


test/ComposableCowPoller.t.sol
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity >=0.8.0 <0.9.0;
+
+import {BaseComposableCoWTest} from "./ComposableCoW.base.t.sol";
+
+import {ComposableCowPoller} from "../src/types/ComposableCowPoller.sol";
```
igorroncevic: nit: why not use absolute path instead of relative? this is nice so imports don't break if something changes.

e.g. `import {ComposableCowPoller} from "src/types/ComposableCowPoller.sol";`

just have to configure base paths in [`foundry.toml`](https://github.com/cowprotocol/solver-7702-delegate/blob/main/foundry.toml#L2)

igorroncevic (dismissed):
nits, but nice and simple overall

igorroncevic:

test/ComposableCowPoller.t.sol
```diff
@@ -0,0 +1,23 @@
...
+    ComposableCowPoller poller;
+
+    function setUp() public virtual override(BaseComposableCoWTest) {
+        super.setUp();
+
+        poller = new ComposableCowPoller();
+    }
+
+    function test_deployment() public {
+        assertTrue(address(poller) != address(0), "poller deployed");
```
kaze-cow: technically this condition only verifies that the constructor returned an address; it doesn't verify that the code was actually saved to the address that was returned

a better check would be:

```suggestion
        assertTrue(address(poller).code.length > 0, "poller deployed");
```

igorroncevic (approved):
LGTM

PR #123, opened 2026-07-10
feat: register a polling schedule

anxolin:
Part of #121. Follow uo #116

The polling requires a pre-authorization from the funding account (the registration).

This pre-authorization is done by registering a **polling schedule**

## What
Adds the struct of the register.

The registrations are done using a `scheduleId` (`bytes32`). The PR provides a `pure` function that generates this id deterministically based on the funder, handler, owner and salt.

This pre-register is what will allow in a follow up PR to implement the constraints in the function that does the actual poll.

## Test plan
```bash
```
Test pass

### DIFF
--- a/src/types/ComposableCowPoller.sol
+++ b/src/types/ComposableCowPoller.sol
@@ -8,15 +8,23 @@ contract ComposableCowPoller {
     /// @notice Parameters for a JIT funding schedule.
     /// @dev A schedule is uniquely identified by its funder, handler, owner, and salt.
     struct Schedule {
-        IConditionalOrderGenerator handler; // the conditional-order handler to poll (e.g. the TWAP type)
-        address funder; // source of funds; the only registrant
-        address owner; // order owner; the pull destination
-        bytes32 salt; // the conditional order's salt
-        bytes staticInput; // the order's staticInput
+        /// @notice The conditional-order handler to poll, such as the TWAP type.
+        IConditionalOrderGenerator handler;
+        /// @notice The address allowed to register this schedule and later debited for sell tokens.
+        /// @dev It can be an EOA or contract and may be the same address as `owner`.
+        address funder;
+        /// @notice The address that owns the ComposableCoW conditional order and receives the pulled funds.
+        /// @dev It can be an EOA or contract and may be the same address as `funder`.
+        address owner;
+        /// @notice A user-controlled namespace for this schedule.
+        /// @dev Use a value unique to the user and order. Deriving it from order-defining static-input values,
+        ///      excluding appData, is recommended.
+        bytes32 salt;
+        /// @notice The static input passed to the handler when it generates an order.
+        bytes staticInput;
     }
 
-    /// @dev Keyed by `id == scheduleId(funder, handler, owner, salt)`, which is independent of the
-    ///      order's `appData`.
+    /// @dev Keyed by `id == scheduleId(schedule)`, which excludes the order's `appData`.
     mapping(bytes32 => Schedule) public schedules;
 
     /// @notice Thrown when someone other than the schedule funder registers or updates a schedule.
@@ -31,17 +39,10 @@ contract ComposableCowPoller {
     /// @notice Computes the deterministic, appData-independent schedule key.
     /// @dev `staticInput` is excluded because its appData can depend on this key.
     ///      Use a different salt for concurrent schedules with the same funder, handler, and owner.
-    /// @param funder The token source.
-    /// @param handler The conditional-order generator.
-    /// @param owner The conditional-order owner.
-    /// @param salt The conditional-order salt.
+    /// @param schedule The schedule whose identity fields determine the key.
     /// @return The schedule key.
-    function scheduleId(address funder, IConditionalOrderGenerator handler, address owner, bytes32 salt)
-        public
-        pure
-        returns (bytes32)
-    {
-        return keccak256(abi.encode(funder, handler, owner, salt));
+    function scheduleId(Schedule memory schedule) public pure returns (bytes32) {
+        return keccak256(abi.encode(schedule.funder, schedule.handler, schedule.owner, schedule.salt));
     }
 
     /// @notice Registers or updates a schedule.
@@ -51,7 +52,7 @@ contract ComposableCowPoller {
     /// @return id The deterministic key of the stored schedule.
     function register(Schedule calldata schedule) external returns (bytes32 id) {
         if (msg.sender != schedule.funder) revert OnlyFunder();
-        id = scheduleId(schedule.funder, schedule.handler, schedule.owner, schedule.salt);
+        id = scheduleId(schedule);
         schedules[id] = schedule;
         emit ScheduleRegistered(id, schedule.owner, schedule.funder);
     }
--- a/src/types/ComposableCowPoller.sol
+++ b/src/types/ComposableCowPoller.sol
@@ -1,5 +1,41 @@
 // SPDX-License-Identifier: GPL-3.0
 pragma solidity >=0.8.0 <0.9.0;
 
+import {IConditionalOrderGenerator} from "../interfaces/IConditionalOrder.sol";
+
 /// @title ComposableCowPoller - Just-in-time funding for composable conditional orders.
-contract ComposableCowPoller {}
+contract ComposableCowPoller {
+    struct Schedule {
+        IConditionalOrderGenerator handler; // the conditional-order handler to poll (e.g. the TWAP type)
+        address funder; // source of funds; the only registrant
+        address owner; // order owner; the pull destination
+        bytes32 salt; // the conditional order's salt
+        bytes staticInput; // the order's staticInput
+    }
+
+    /// @dev Keyed by `id == scheduleId(funder, handler, owner, salt)`, which is independent of the
+    ///      order's `appData`.
+    mapping(bytes32 => Schedule) public schedules;
+
+    error OnlyFunder();
+
+    event ScheduleRegistered(bytes32 indexed id, address indexed owner, address indexed funder);
+
+    /// @notice Deterministic, appData-independent schedule key.
+    function scheduleId(address funder, IConditionalOrderGenerator handler, address owner, bytes32 salt)
+        public
+        pure
+        returns (bytes32)
+    {
+        return keccak256(abi.encode(funder, handler, owner, salt));
+    }
+
+    /// @notice Register or update a schedule. Only the funds source may register, and the `id` is
+    ///         namespaced by `funder`.
+    function register(Schedule calldata schedule) external returns (bytes32 id) {
+        if (msg.sender != schedule.funder) revert OnlyFunder();
+        id = scheduleId(schedule.funder, schedule.handler, schedule.owner, schedule.salt);
+        schedules[id] = schedule;
+        emit ScheduleRegistered(id, schedule.owner, schedule.funder);
+    }
+}
--- a/src/types/ComposableCowPoller.sol
+++ b/src/types/ComposableCowPoller.sol
@@ -1,10 +1,12 @@
 // SPDX-License-Identifier: GPL-3.0
 pragma solidity >=0.8.0 <0.9.0;
 
-import {IConditionalOrderGenerator} from "../interfaces/IConditionalOrder.sol";
+import {IConditionalOrderGenerator} from "src/interfaces/IConditionalOrder.sol";
 
 /// @title ComposableCowPoller - Just-in-time funding for composable conditional orders.
 contract ComposableCowPoller {
+    /// @notice Parameters for a JIT funding schedule.
+    /// @dev A schedule is uniquely identified by its funder, handler, owner, and salt.
     struct Schedule {
         IConditionalOrderGenerator handler; // the conditional-order handler to poll (e.g. the TWAP type)
         address funder; // source of funds; the only registrant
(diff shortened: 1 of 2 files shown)


src/types/ComposableCowPoller.sol
```diff
@@ -1,5 +1,41 @@
 // SPDX-License-Identifier: GPL-3.0
 pragma solidity >=0.8.0 <0.9.0;
 
+import {IConditionalOrderGenerator} from "../interfaces/IConditionalOrder.sol";
```
igorroncevic: nit: absolute import path?

src/types/ComposableCowPoller.sol
```diff
@@ -1,5 +1,41 @@
...
+
 /// @title ComposableCowPoller - Just-in-time funding for composable conditional orders.
-contract ComposableCowPoller {}
+contract ComposableCowPoller {
```
igorroncevic, line 7: nit: missing a lot of NatSpec, think we should add it?

src/types/ComposableCowPoller.sol
```diff
@@ -1,5 +1,41 @@
...
+    {
+        return keccak256(abi.encode(funder, handler, owner, salt));
+    }
+
+    /// @notice Register or update a schedule. Only the funds source may register, and the `id` is
+    ///         namespaced by `funder`.
+    function register(Schedule calldata schedule) external returns (bytes32 id) {
+        if (msg.sender != schedule.funder) revert OnlyFunder();
+        id = scheduleId(schedule.funder, schedule.handler, schedule.owner, schedule.salt);
+        schedules[id] = schedule;
```
igorroncevic, line 56: The same funder can register the same `handler`/`owner`/`salt` with different `staticInput` and replace the previous schedule. 

Because `staticInput` can be anything and we can't use it directly because it contains `appData` (which comes with circular dependency), we can't rely on it improving the uniqueness of the ID.

The only thing allowing user to have multiple e.g. TWAP orders is a different salt, so ideally, we'd document this as a very important flow detail.

test/ComposableCowPoller.t.sol
```diff
@@ -1,23 +1,133 @@
...
+        // The owner (safe1) starts with no sell token: funds arrive just-in-time.
+        deal(address(token0), address(safe1), 0);
+    }
+
+    function _bundle() internal view returns (TWAPOrder.Data memory) {
+        return
+            TWAPOrder.Data({
+                sellToken: token0,
+                buyToken: token1,
+                receiver: address(0), // the owner itself
```
igorroncevic: nit: is this documented behavior?

test/ComposableCowPoller.t.sol
```diff
@@ -1,23 +1,133 @@
...
+                partSellAmount: PART,
+                minPartLimit: LIMIT,
+                t0: 0, // resolved from the cabinet via createWithContext
+                n: N,
+                t: FREQ,
+                span: 0, // valid for the whole epoch
```
igorroncevic: which epoch? is `epoch = (t0, t0 + N * FREQ)`?

src/types/ComposableCowPoller.sol
```diff
@@ -1,5 +1,58 @@
...
+import {IConditionalOrderGenerator} from "src/interfaces/IConditionalOrder.sol";
+
 /// @title ComposableCowPoller - Just-in-time funding for composable conditional orders.
-contract ComposableCowPoller {}
+contract ComposableCowPoller {
+    /// @notice Parameters for a JIT funding schedule.
+    /// @dev A schedule is uniquely identified by its funder, handler, owner, and salt.
+    struct Schedule {
+        IConditionalOrderGenerator handler; // the conditional-order handler to poll (e.g. the TWAP type)
+        address funder; // source of funds; the only registrant
```
kaze-cow: based solely on this comment, its unclear: in the case of polling for a TWAP order using CoW Shed and an EOA, is the `funder` the EOA, or the `CoWShed`?

src/types/ComposableCowPoller.sol
```diff
@@ -1,5 +1,58 @@
...
+        address owner; // order owner; the pull destination
```
kaze-cow: from this comment I assume it is the CoWShed. I would rename this field to `orderOwner` and clarify in the comment that this is the CoW Settlement order, as opposed to any other kind of order that may become relevant.

src/types/ComposableCowPoller.sol
```diff
@@ -1,5 +1,58 @@
...
+        bytes32 salt; // the conditional order's salt
+        bytes staticInput; // the order's staticInput
```
kaze-cow: comments are kind of redundant. maybe

```suggestion
        /// Used to set a new poller schedule instance with the same parameters
        bytes32 salt;
        /// The `staticInput` parameter used for `getTradableOrder` when calling `handler`
        bytes staticInput;
```

src/types/ComposableCowPoller.sol
```diff
@@ -1,5 +1,58 @@
...
 pragma solidity >=0.8.0 <0.9.0;
 
+import {IConditionalOrderGenerator} from "src/interfaces/IConditionalOrder.sol";
+
 /// @title ComposableCowPoller - Just-in-time funding for composable conditional orders.
-contract ComposableCowPoller {}
+contract ComposableCowPoller {
+    /// @notice Parameters for a JIT funding schedule.
+    /// @dev A schedule is uniquely identified by its funder, handler, owner, and salt.
+    struct Schedule {
```
kaze-cow, line 10: all of the fields in this struct comments could be docstrings (`///` before the field)

src/types/ComposableCowPoller.sol
```diff
@@ -1,5 +1,58 @@
...
+    /// @param owner The conditional-order owner.
+    /// @param salt The conditional-order salt.
+    /// @return The schedule key.
+    function scheduleId(address funder, IConditionalOrderGenerator handler, address owner, bytes32 salt)
+        public
+        pure
+        returns (bytes32)
+    {
+        return keccak256(abi.encode(funder, handler, owner, salt));
+    }
```
kaze-cow, line 46: Even though `staticInput` isnt used in the schedule ID, it still would make the most sense here to simply take in the `Schedule` struct to make it simpler for the front end. Also, its wierd that the order of `address funder` and `IConditionalOrderGenerator handler` have been flipped from the original structure, and this could be problematic on the frontend because the type is ultimately `address` in both cases.

src/types/ComposableCowPoller.sol
```diff
@@ -1,5 +1,58 @@
...
+
+    /// @notice Registers or updates a schedule.
+    /// @dev Registering the same funder, handler, owner, and salt replaces the stored schedule.
+    ///      Only the funds source may register, and the ID is namespaced by the funder.
+    /// @param schedule The schedule to store.
+    /// @return id The deterministic key of the stored schedule.
+    function register(Schedule calldata schedule) external returns (bytes32 id) {
+        if (msg.sender != schedule.funder) revert OnlyFunder();
```
kaze-cow: its now possible to do

```suggestion
    require(msg.sender == schedule.funder, OnlyFunder());
```

src/types/ComposableCowPoller.sol
```diff
@@ -1,5 +1,58 @@
...
+    mapping(bytes32 => Schedule) public schedules;
+
+    /// @notice Thrown when someone other than the schedule funder registers or updates a schedule.
+    error OnlyFunder();
+
+    /// @notice Emitted when a schedule is registered or updated.
+    /// @param id The deterministic key of the schedule.
+    /// @param owner The conditional-order owner and pull destination.
+    /// @param funder The token source that registered the schedule.
+    event ScheduleRegistered(bytes32 indexed id, address indexed owner, address indexed funder);
```
kaze-cow, line 37: for completeness it would be good to include all fields of `Schedule` in the log

test/ComposableCowPoller.t.sol
```diff
@@ -1,23 +1,155 @@
...
+import {ComposableCowPoller} from "src/types/ComposableCowPoller.sol";
+import {CurrentBlockTimestampFactory} from "src/value_factories/CurrentBlockTimestampFactory.sol";
+import {IConditionalOrderGenerator} from "src/interfaces/IConditionalOrder.sol";
 
 /// @title ComposableCowPoller unit tests
-/// @notice Base scaffolding for the poller. Feature-specific tests (register / revoke / topUp) are
-///         added in follow-up PRs.
+/// @notice Exercises registering a schedule for a composable TWAP created via `createWithContext`.
 contract ComposableCowPollerTest is BaseComposableCoWTest {
+    uint256 constant PART = 100e18;
```
kaze-cow: this is the amount to trade on each TWAP trade, right? The constant (or at least the naming) actually makes it harder to understand what is going on, so it seems like it would be best to rename this to something like `TWAP_PART_AMOUNT`, or perhaps remove it entirely.

test/ComposableCowPoller.t.sol
```diff
@@ -1,23 +1,155 @@
...
+    uint256 constant LIMIT = 1e18;
```
kaze-cow, line 16: this value is only used in the `_bundle` function so low value in having it as a constant here.

kaze-cow (dismissed):
there is probably some shady security issue with simply not including the `staticInput` in the schedule ID, but I need to see the rest of the code put together before I will comment further.

appears there are currently merge conflicts that need fixing

src/types/ComposableCowPoller.sol
igorroncevic: While I prefer this style of reverts, this is not a standard adopted in any CoW repo IIRC.

src/types/ComposableCowPoller.sol
igorroncevic: The current fields are the most important ones, and adding more could just spike up gas costs at not much additional benefits (e.g. staticInput, handler etc)

test/ComposableCowPoller.t.sol
igorroncevic: Hardcoded magic values are really hard to understand, so I'd keep it as-is

src/types/ComposableCowPoller.sol
igorroncevic: IIUC, funder can be anyone that approves the Poller contract

src/types/ComposableCowPoller.sol
igorroncevic: I don't think this brings any value. `owner` can be anyone that will receive funds in the end

src/types/ComposableCowPoller.sol
anxolin: I believe igor added better documentation. Resolving, let me know if good enough

src/types/ComposableCowPoller.sol
anxolin: Igor added additional comments

```
/// @dev It can be an EOA or contract and may be the same address as `funder`.
address owner;
```

hopefully is not confusing any more. Also owner is the notation in composable cow for the owner of the order. Closing, but feel free to reopen

src/types/ComposableCowPoller.sol
anxolin: Fair enough, I see Igor addressed this comment, and I leave the changes. 

I'm not a fan of positional arguments cause they are error prone, reason I added them here was so we are more explicit, and giving there's few, though would be ok. Passing the struct hides the fact that half of the data from it is not participating in the id generation.

test/ComposableCowPoller.t.sol
anxolin: addressed by igor

src/types/ComposableCowPoller.sol
anxolin: I think is a good idea to try to include them. Specially if we can override schedules or in case it help an indexer map it to the specific order and gas is not too much, but will experiment as an interation https://github.com/cowprotocol/composable-cow/pull/129

src/types/ComposableCowPoller.sol
```diff
@@ -1,5 +1,59 @@
...
+    function scheduleId(Schedule memory schedule) public pure returns (bytes32) {
+        return keccak256(abi.encode(schedule.funder, schedule.handler, schedule.owner, schedule.salt));
+    }
+
+    /// @notice Registers or updates a schedule.
+    /// @dev Registering the same funder, handler, owner, and salt replaces the stored schedule.
+    ///      Only the funds source may register, and the ID is namespaced by the funder.
+    /// @param schedule The schedule to store.
+    /// @return id The deterministic key of the stored schedule.
+    function register(Schedule calldata schedule) external returns (bytes32 id) {
```
igorroncevic, line 53: owner could be `address(0)`, which could render funds unusable if they ended up there. handler being `address(0)` could be annoying to handle off-chain but isn't as big of an issue. I suggest we reject such cases.

anxolin: That's a great suggestion, but I don't think its relevant in this case. 

The reason is because the polling in #125 only polls if the composable cow order is authorised. 
Authorization is a call to `createOrder` in composable cow, which uses `msg.sender` which can't be 0x00.

So, yes, even if you register the wrong handler or owner, the transfer can't be done. I added a note on #125 (because there is where we add the polling): a553c4c6c48e9ade4094ac7921fed3f4d3f269f8 

I personally don't care too much if someone can register owners/handlers that don't make sense. Source of truth will be the composable cow authorization

igorroncevic: I would personally always be on the safe side and add this small check, but I won't block the PR as it is not critical and the safety of this check is brought through another piece of code.

igorroncevic (approved):

PR #124, opened 2026-07-10
feat: revoke a polling schedule

anxolin:
Part of #121. Follows up https://github.com/cowprotocol/composable-cow/pull/123

## What
Adds `revoke(bytes32 id)` — only the schedule's `funder` may revoke, which clears the stored schedule and emits `ScheduleRevoked`.

## Why
Lets a funder tear down a schedule they registered. (A standing ERC-20 allowance should be revoked separately.)

## Test plan
```bash
```
Covers `test_revoke_clearsSchedule` (funder-only guard + schedule cleared).

### DIFF
--- a/src/types/ComposableCowPoller.sol
+++ b/src/types/ComposableCowPoller.sol
@@ -35,6 +35,7 @@ contract ComposableCowPoller {
     /// @param owner The conditional-order owner and pull destination.
     /// @param funder The token source that registered the schedule.
     event ScheduleRegistered(bytes32 indexed id, address indexed owner, address indexed funder);
+    event ScheduleRevoked(bytes32 indexed id);
 
     /// @notice Computes the deterministic, appData-independent schedule key.
     /// @dev `staticInput` is excluded because its appData can depend on this key.
@@ -55,5 +56,13 @@ contract ComposableCowPoller {
         id = scheduleId(schedule);
         schedules[id] = schedule;
         emit ScheduleRegistered(id, schedule.owner, schedule.funder);
+    }
+
+    /// @notice Revoke a schedule. Only the funds source may do so. A standing ERC-20 allowance
+    ///         should be revoked separately to fully close the surface.
+    function revoke(bytes32 id) external {
+        if (msg.sender != schedules[id].funder) revert OnlyFunder();
+        delete schedules[id];
+        emit ScheduleRevoked(id);
     }
 }
--- a/src/types/ComposableCowPoller.sol
+++ b/src/types/ComposableCowPoller.sol
@@ -35,7 +35,12 @@ contract ComposableCowPoller {
     /// @param owner The conditional-order owner and pull destination.
     /// @param funder The token source that registered the schedule.
     event ScheduleRegistered(bytes32 indexed id, address indexed owner, address indexed funder);
-    event ScheduleRevoked(bytes32 indexed id);
+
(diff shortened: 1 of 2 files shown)


src/types/ComposableCowPoller.sol
```diff
@@ -20,6 +20,7 @@ contract ComposableCowPoller {
     error OnlyFunder();
 
     event ScheduleRegistered(bytes32 indexed id, address indexed owner, address indexed funder);
+    event ScheduleRevoked(bytes32 indexed id);
```
igorroncevic: Should we emit owner or the funder, so we can index their revoked transactions?

igorroncevic (approved):
only one nit, LGTM otherwise

src/types/ComposableCowPoller.sol
```diff
@@ -38,4 +44,13 @@ contract ComposableCowPoller {
         schedules[id] = schedule;
         emit ScheduleRegistered(id, schedule.owner, schedule.funder);
     }
+
+    /// @notice Revoke a schedule. Only the funds source may do so. A standing ERC-20 allowance
+    ///         should be revoked separately to fully close the surface.
+    function revoke(bytes32 id) external {
+        Schedule storage schedule = schedules[id];
+        if (msg.sender != schedule.funder) revert OnlyFunder();
```
kaze-cow, line 70: it is now possible/simpler to do

```suggestion
    require(msg.sender == schedule.funder, OnlyFunder());
```

test/ComposableCowPoller.t.sol
```diff
@@ -130,4 +132,27 @@ contract ComposableCowPollerTest is BaseComposableCoWTest {
...
+
+        vm.expectEmit(true, true, true, true, address(poller));
+        emit ScheduleRevoked(id, address(safe1), funder);
+
+        vm.prank(funder);
+        poller.revoke(id);
+
+        // The schedule is cleared: its funder is zeroed.
+        (, address scheduleFunder, , ,) = poller.schedules(id);
+        assertEq(scheduleFunder, address(0), "schedule cleared");
```
kaze-cow: it would be most complete to validate that every field here has been zeroed out

src/types/ComposableCowPoller.sol
igorroncevic: While I prefer this style of reverts, this is not a standard adopted in any CoW repo IIRC.

anxolin:
Thank for the review, and thanks @igorroncevic for the fixes on top of the PR

src/types/ComposableCowPoller.sol
```diff
@@ -56,4 +62,13 @@ contract ComposableCowPoller {
         schedules[id] = schedule;
         emit ScheduleRegistered(id, schedule.owner, schedule.funder);
     }
+
+    /// @notice Revoke a schedule. Only the funds source may do so. A standing ERC-20 allowance
+    ///         should be revoked separately to fully close the surface.
+    function revoke(bytes32 id) external {
+        Schedule storage schedule = schedules[id];
+    require(msg.sender == schedule.funder, OnlyFunder());
```
igorroncevic: nit: messed up whitespace

igorroncevic (dismissed):

kaze-cow (approved):

igorroncevic (approved):

PR #125, opened 2026-07-10
feat: pollFunds for just-in-time order funding

anxolin:
Part of #121. Follows on https://github.com/cowprotocol/composable-cow/pull/124

## Note on the merge / approvals
This PR is merged with 2 approvals, even though it doesnt show. Apparently because of the settings in this repo, when I merge and squashed a dependent PR it dismissed the 2 approvals:
```
{at: 19:32:32, was: "approved", reason: "The base branch was changed.", by: igorroncevic}
{at: 19:32:32, was: "approved", reason: "The base branch was changed.", by: kaze-cow}
S
```
However, the commits are the exact ones that were approved, so merged as it is. 

## What
This is the PR that adding the main feature.

Adds the just-in-time funding entry point (`pollFunds`). This is the function that will be called by the hooks to poll the funds.

The method will do a series of validations:
- The schedule was registered,
- The polling hasn't been done yet for the current order. For checking this, it will check with the pre-registered handler to generate an order, and from it, derices the `orderUid`. This is how we can reuse the logic from the handlers to be used for the polling too. For example, in TWAP makes sure we don't pull twice for the same leg.
- It also checks the composable cow order has been authorised . This way, cancellations of composable cow orders will also work out of the box. 

## Topup vs pollFunds
Initially in my first draft I implemented a method to top up the account (adding whatever balance that is missing in order to be able to fullfull the `order.sellAmount`.

I decided to rename to `pollFunds` to move the funds unconditionally (don't check the balance. This way, we can reuse the same trader contract for more concurrent orders, and we don't run into race conditions where some order uses the funds of some other.

## Test plan
```bash
```

Test should pass

### DIFF
--- a/src/types/ComposableCowPoller.sol
+++ b/src/types/ComposableCowPoller.sol
@@ -1,10 +1,15 @@
 // SPDX-License-Identifier: GPL-3.0
 pragma solidity >=0.8.0 <0.9.0;
 
-import {IConditionalOrderGenerator} from "src/interfaces/IConditionalOrder.sol";
+import {IERC20, GPv2Order} from "cowprotocol/contracts/libraries/GPv2Order.sol";
+
+import {ComposableCoW} from "src/ComposableCoW.sol";
+import {IConditionalOrder, IConditionalOrderGenerator} from "src/interfaces/IConditionalOrder.sol";
 
 /// @title ComposableCowPoller - Just-in-time funding for composable conditional orders.
 contract ComposableCowPoller {
+    ComposableCoW public immutable composableCow;
+
     /// @notice Parameters for a JIT funding schedule.
     /// @dev A schedule is uniquely identified by its funder, handler, owner, and salt.
     struct Schedule {
@@ -27,14 +32,24 @@ contract ComposableCowPoller {
     /// @dev Keyed by `id == scheduleId(schedule)`, which excludes the order's `appData`.
     mapping(bytes32 => Schedule) public schedules;
 
+    /// @dev `id => digest` of the order last funded, so each order is funded at most once.
+    mapping(bytes32 => bytes32) public lastFunded;
+
     /// @notice Thrown when someone other than the schedule funder registers or updates a schedule.
     error OnlyFunder();
+    error NoSchedule();
+    error OrderNotLive();
 
     /// @notice Emitted when a schedule is registered or updated.
     /// @param id The deterministic key of the schedule.
     /// @param owner The conditional-order owner and pull destination.
     /// @param funder The token source that registered the schedule.
     event ScheduleRegistered(bytes32 indexed id, address indexed owner, address indexed funder);
+    event Pulled(bytes32 indexed id, bytes32 orderDigest, uint256 amount);
+
+    constructor(ComposableCoW _composableCow) {
+        composableCow = _composableCow;
+    }
 
     /// @notice Emitted when a schedule is revoked.
     /// @param id The deterministic key of the revoked schedule.
@@ -60,6 +75,7 @@ contract ComposableCowPoller {
         if (msg.sender != schedule.funder) revert OnlyFunder();
         id = scheduleId(schedule);
         schedules[id] = schedule;
+        delete lastFunded[id]; // reset funding history
         emit ScheduleRegistered(id, schedule.owner, schedule.funder);
     }
 
@@ -70,5 +86,37 @@ contract ComposableCowPoller {
         if (msg.sender != schedule.funder) revert OnlyFunder();
         emit ScheduleRevoked(id, schedule.owner, schedule.funder);
         delete schedules[id];
+    }
+
+    /// @notice Move the current order's `sellAmount` from the funder to the owner. Permissionless.
+    ///         The full amount always moves (no balance check), so one owner can serve several
+    ///         concurrent orders.
+    function pollFunds(bytes32 id) external {
+        Schedule memory schedule = schedules[id];
+        if (schedule.funder == address(0)) revert NoSchedule();
+
+        // Re-derive `ctx` on-chain, so `pollFunds(id)` stays independent of the order's `appData`.
+        bytes32 ctx = composableCow.hash(
+            IConditionalOrder.ConditionalOrderParams({
+                handler: schedule.handler,
+                salt: schedule.salt,
+                staticInput: schedule.staticInput
+            })
+        );
+
+        // The order must still be authorised; `remove` disables the poller.
+        if (!composableCow.singleOrders(schedule.owner, ctx)) revert OrderNotLive();
+
+        // The handler yields the current order and reverts outside its window.
+        GPv2Order.Data memory order =
+            schedule.handler.getTradeableOrder(schedule.owner, address(this), ctx, schedule.staticInput, bytes(""));
+
+        // Fund each order at most once.
+        bytes32 digest = GPv2Order.hash(order, composableCow.domainSeparator());
+        if (digest == lastFunded[id]) return;
+        lastFunded[id] = digest;
+
+        order.sellToken.transferFrom(schedule.funder, schedule.owner, order.sellAmount);
+        emit Pulled(id, digest, order.sellAmount);
     }
 }
--- a/src/types/ComposableCowPoller.sol
+++ b/src/types/ComposableCowPoller.sol
@@ -11,7 +11,8 @@ import {IConditionalOrder, IConditionalOrderGenerator} from "src/interfaces/ICon
 contract ComposableCowPoller {
     using GPv2SafeERC20 for IERC20;
 
-    ComposableCoW public immutable composableCow;
+    /// @dev `ComposableCoW` stores the settlement domain separator supplied at deployment.
+    ComposableCoW public immutable COMPOSABLE_COW;
 
     /// @notice Parameters for a JIT funding schedule.
     /// @dev A schedule is uniquely identified by its funder, handler, owner, and salt.
@@ -25,8 +26,8 @@ contract ComposableCowPoller {
         /// @dev It can be an EOA or contract and may be the same address as `funder`.
         address owner;
         /// @notice A user-controlled namespace for this schedule.
-        /// @dev Use a value unique to the user and order. Deriving it from order-defining static-input values,
-        ///      excluding appData, is recommended.
+        /// @dev Keep it unique to the user and logical order. A good default is the hash of the
+        ///      order-defining static-input values with any appData field set to zero.
         bytes32 salt;
         /// @notice The static input passed to the handler when it generates an order.
         bytes staticInput;
@@ -35,8 +36,8 @@ contract ComposableCowPoller {
     /// @dev Keyed by `id == scheduleId(schedule)`, which excludes the order's `appData`.
     mapping(bytes32 => Schedule) public schedules;
 
-    /// @dev `id => digest` of the order last funded, so each order is funded at most once.
-    mapping(bytes32 => bytes32) public lastFunded;
+    /// @dev `id => digest => funded`. History survives schedule updates so an old order cannot be replayed.
+    mapping(bytes32 => mapping(bytes32 => bool)) public funded;
 
     /// @notice Thrown when someone other than the schedule funder registers or updates a schedule.
     error OnlyFunder();
@@ -51,7 +52,7 @@ contract ComposableCowPoller {
     event Pulled(bytes32 indexed id, bytes32 orderDigest, uint256 amount);
 
     constructor(ComposableCoW _composableCow) {
-        composableCow = _composableCow;
+        COMPOSABLE_COW = _composableCow;
     }
 
     /// @notice Emitted when a schedule is revoked.
@@ -62,7 +63,8 @@ contract ComposableCowPoller {
 
     /// @notice Computes the deterministic, appData-independent schedule key.
     /// @dev `staticInput` is excluded because its appData can depend on this key.
-    ///      Use a different salt for concurrent schedules with the same funder, handler, and owner.
+    ///      Keep the salt unique to the user and logical order. A good default is the hash of the
+    ///      order-defining static-input values with any appData field set to zero.
     /// @param schedule The schedule whose identity fields determine the key.
     /// @return The schedule key.
     function scheduleId(Schedule memory schedule) public pure returns (bytes32) {
@@ -71,14 +73,14 @@ contract ComposableCowPoller {
 
     /// @notice Registers or updates a schedule.
     /// @dev Registering the same funder, handler, owner, and salt replaces the stored schedule.
-    ///      Only the funds source may register, and the ID is namespaced by the funder.
+    ///      Only the funds source may register, and the ID is namespaced by the funder. Funding
+    ///      history is preserved across updates; use a new salt for a new logical schedule.
     /// @param schedule The schedule to store.
     /// @return id The deterministic key of the stored schedule.
     function register(Schedule calldata schedule) external returns (bytes32 id) {
         if (msg.sender != schedule.funder) revert OnlyFunder();
         id = scheduleId(schedule);
         schedules[id] = schedule;
-        delete lastFunded[id]; // reset funding history
         emit ScheduleRegistered(id, schedule.owner, schedule.funder);
     }
 
@@ -99,25 +101,23 @@ contract ComposableCowPoller {
         if (schedule.funder == address(0)) revert NoSchedule();
 
         // Re-derive `ctx` on-chain, so `pollFunds(id)` stays independent of the order's `appData`.
-        bytes32 ctx = composableCow.hash(
+        bytes32 ctx = COMPOSABLE_COW.hash(
             IConditionalOrder.ConditionalOrderParams({
-                handler: schedule.handler,
-                salt: schedule.salt,
-                staticInput: schedule.staticInput
+                handler: schedule.handler, salt: schedule.salt, staticInput: schedule.staticInput
             })
         );
 
         // The order must still be authorised; `remove` disables the poller.
-        if (!composableCow.singleOrders(schedule.owner, ctx)) revert OrderNotLive();
+        if (!COMPOSABLE_COW.singleOrders(schedule.owner, ctx)) revert OrderNotLive();
 
         // The handler yields the current order and reverts outside its window.
         GPv2Order.Data memory order =
             schedule.handler.getTradeableOrder(schedule.owner, address(this), ctx, schedule.staticInput, bytes(""));
 
-        // Fund each order at most once.
-        bytes32 digest = GPv2Order.hash(order, composableCow.domainSeparator());
-        if (digest == lastFunded[id]) return;
-        lastFunded[id] = digest;
+        // `ComposableCoW` exposes the settlement domain separator it received at deployment.
+        bytes32 digest = GPv2Order.hash(order, COMPOSABLE_COW.domainSeparator());
+        if (funded[id][digest]) return;
+        funded[id][digest] = true;
 
         order.sellToken.safeTransferFrom(schedule.funder, schedule.owner, order.sellAmount);
         emit Pulled(id, digest, order.sellAmount);
--- a/src/types/ComposableCowPoller.sol
+++ b/src/types/ComposableCowPoller.sol
@@ -2,12 +2,15 @@
 pragma solidity >=0.8.0 <0.9.0;
 
 import {IERC20, GPv2Order} from "cowprotocol/contracts/libraries/GPv2Order.sol";
+import {GPv2SafeERC20} from "cowprotocol/contracts/libraries/GPv2SafeERC20.sol";
 
 import {ComposableCoW} from "src/ComposableCoW.sol";
 import {IConditionalOrder, IConditionalOrderGenerator} from "src/interfaces/IConditionalOrder.sol";
 
 /// @title ComposableCowPoller - Just-in-time funding for composable conditional orders.
 contract ComposableCowPoller {
+    using GPv2SafeERC20 for IERC20;
+
     ComposableCoW public immutable composableCow;
 
     /// @notice Parameters for a JIT funding schedule.
@@ -116,7 +119,7 @@ contract ComposableCowPoller {
         if (digest == lastFunded[id]) return;
         lastFunded[id] = digest;
 
-        order.sellToken.transferFrom(schedule.funder, schedule.owner, order.sellAmount);
+        order.sellToken.safeTransferFrom(schedule.funder, schedule.owner, order.sellAmount);
         emit Pulled(id, digest, order.sellAmount);
     }
 }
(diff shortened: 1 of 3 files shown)


src/types/ComposableCowPoller.sol
```diff
@@ -1,10 +1,15 @@
 // SPDX-License-Identifier: GPL-3.0
 pragma solidity >=0.8.0 <0.9.0;
 
-import {IConditionalOrderGenerator} from "../interfaces/IConditionalOrder.sol";
+import {IERC20, GPv2Order} from "cowprotocol/contracts/libraries/GPv2Order.sol";
+
+import {ComposableCoW} from "../ComposableCoW.sol";
+import {IConditionalOrder, IConditionalOrderGenerator} from "../interfaces/IConditionalOrder.sol";
```
igorroncevic: nit: absolute import paths?

src/types/ComposableCowPoller.sol
```diff
@@ -36,6 +51,39 @@ contract ComposableCowPoller {
...
+        // The handler yields the current order and reverts outside its window.
+        GPv2Order.Data memory order =
+            schedule.handler.getTradeableOrder(schedule.owner, address(this), ctx, schedule.staticInput, bytes(""));
+
+        // Fund each order at most once.
+        bytes32 digest = GPv2Order.hash(order, composableCow.domainSeparator());
+        if (digest == lastFunded[id]) return;
+        lastFunded[id] = digest;
+
+        order.sellToken.transferFrom(schedule.funder, schedule.owner, order.sellAmount);
```
igorroncevic: Return value is being ignored. This is a problem because some tokens may return `false` without reverting. `lastFunded[id]` is updated, leg is marked as funded, and `Pulled` is emitted, even though no tokens moved.

Use a safe transfer helper that reverts on a `false` return.

src/types/ComposableCowPoller.sol
```diff
@@ -36,6 +51,39 @@ contract ComposableCowPoller {
         if (msg.sender != schedule.funder) revert OnlyFunder();
         id = scheduleId(schedule.funder, schedule.handler, schedule.owner, schedule.salt);
         schedules[id] = schedule;
+        delete lastFunded[id]; // reset funding history
```
igorroncevic: Every successful registration clears `lastFunded[id]`. 

A realistic sequence is:

1. Alice creates a 10-day TWAP and gives the poller approval for 100 USDC.
2. On day 3, the poller pulls the intended 10 USDC into Alice's `CowShed` and records that pull.
3. Alice signs and sends another `register` transaction with the same ID.
4. The contract accepts it and runs `delete lastFunded[id]`.
5. A call to `pollFunds` transfers another 10 USDC into `CowShed`, because it doesn't know about the old order anymore.

This is only weird UX, not an issue, but something to be documented, perhaps even reject overriding existing IDs to not disrupt ongoing orders, unless this was done for a reason.

src/types/ComposableCowPoller.sol
```diff
@@ -36,6 +54,39 @@ contract ComposableCowPoller {
...
+        bytes32 ctx = composableCow.hash(
+            IConditionalOrder.ConditionalOrderParams({
+                handler: schedule.handler,
+                salt: schedule.salt,
+                staticInput: schedule.staticInput
+            })
+        );
+
+        // The order must still be authorised; `remove` disables the poller.
+        if (!composableCow.singleOrders(schedule.owner, ctx)) revert OrderNotLive();
```
kaze-cow:
```suggestion
        require(composableCow.singleOrders(schedule.owner, ctx), OrderNotLive());
```

src/types/ComposableCowPoller.sol
```diff
@@ -1,10 +1,18 @@
...
+import {GPv2SafeERC20} from "cowprotocol/contracts/libraries/GPv2SafeERC20.sol";
+
+import {ComposableCoW} from "src/ComposableCoW.sol";
+import {IConditionalOrder, IConditionalOrderGenerator} from "src/interfaces/IConditionalOrder.sol";
 
 /// @title ComposableCowPoller - Just-in-time funding for composable conditional orders.
 contract ComposableCowPoller {
+    using GPv2SafeERC20 for IERC20;
+
+    ComposableCoW public immutable composableCow;
```
kaze-cow: This should be `COMPOSABLE_COW` due to solidity conventions of immutables

src/types/ComposableCowPoller.sol
kaze-cow: technically its also possible that the order may be authorized through merkle hash, though I assume this functionality just hasn't been used so its ok to ignore this.

src/types/ComposableCowPoller.sol
```diff
@@ -36,6 +54,39 @@ contract ComposableCowPoller {
...
+
+        // The order must still be authorised; `remove` disables the poller.
+        if (!composableCow.singleOrders(schedule.owner, ctx)) revert OrderNotLive();
+
+        // The handler yields the current order and reverts outside its window.
+        GPv2Order.Data memory order =
+            schedule.handler.getTradeableOrder(schedule.owner, address(this), ctx, schedule.staticInput, bytes(""));
+
+        // Fund each order at most once.
+        bytes32 digest = GPv2Order.hash(order, composableCow.domainSeparator());
```
kaze-cow: its weird here to use the domain separator of `composableCow`. Why not use the domain hash of the settlement contract instead? It can be included as an additional immutable property. That way, the order hash will match up with the one actually on the settlement contract.

src/types/ComposableCowPoller.sol
```diff
@@ -17,9 +25,19 @@ contract ComposableCowPoller {
     ///      order's `appData`.
     mapping(bytes32 => Schedule) public schedules;
 
+    /// @dev `id => digest` of the order last funded, so each order is funded at most once.
+    mapping(bytes32 => bytes32) public lastFunded;
```
kaze-cow: with our current pragmatic orders, I don't believe there is any reason why this strategy of only tracking the order last funded would be problematic. However, if a pragmatic order's next order could somehow be controlled or set by a value/field that a solver can manipulate, a single order could be reused/pulled multiple times by switching back and forth between two valid `getTradableOrders`

test/ComposableCowPoller.t.sol
```diff
@@ -130,4 +136,180 @@ contract ComposableCowPollerTest is BaseComposableCoWTest {
...
+            token0.balanceOf(address(safe1)),
+            PART,
+            "owner funded with exactly one part"
+        );
+        assertEq(
+            token0.balanceOf(funder),
+            PART * N - PART,
+            "exactly one part left the funder"
+        );
+    }
```
kaze-cow: this test is superfluous with the following `test_pollFunds_movesFullAmountUnconditionally`

src/types/ComposableCowPoller.sol
```diff
@@ -36,6 +54,39 @@ contract ComposableCowPoller {
...
         id = scheduleId(schedule.funder, schedule.handler, schedule.owner, schedule.salt);
         schedules[id] = schedule;
+        delete lastFunded[id]; // reset funding history
         emit ScheduleRegistered(id, schedule.owner, schedule.funder);
     }
+
+    /// @notice Move the current order's `sellAmount` from the funder to the owner. Permissionless.
+    ///         The full amount always moves (no balance check), so one owner can serve several
+    ///         concurrent orders.
+    function pollFunds(bytes32 id) external {
```
kaze-cow, line 112: this function could return `false` or `true` depending upon whether funds were ultimately transferred or not. This would mostly make the tests a bit more clear and offer a bit more surface in the off chance that a smart contract wants to use this contract and know if funds were transferred or not without relying on balance checks.

test/ComposableCowPoller.t.sol
```diff
@@ -130,4 +136,180 @@ contract ComposableCowPollerTest is BaseComposableCoWTest {
...
+                address(safe1),
+                PART
+            ),
+            abi.encode(false)
+        );
+
+        vm.expectRevert(bytes("GPv2: failed transferFrom"));
+        poller.pollFunds(id);
+
+        assertEq(poller.lastFunded(id), bytes32(0), "failed pull is not recorded");
```
kaze-cow: technically this check is superfluous if the transaction is already known to have reverted

test/ComposableCowPoller.t.sol
```diff
@@ -130,4 +136,180 @@ contract ComposableCowPollerTest is BaseComposableCoWTest {
...
+                "one part funded per window"
+            );
+        }
+    }
+
+    /// @dev The anti-premature-execution guard: once a part is funded, the *next* part's funds
+    ///      cannot be pulled until time advances into its window — even after the part settles and
+    ///      drains the owner. Without the per-order guard, the drained owner would be refilled
+    ///      immediately and that balance would be sold as the next part, a full interval early.
+    function test_pollFunds_cannotFundFuturePartEarly() public {
```
kaze-cow: this test is superfluous with `test_pollFunds_idempotentWithinPart`. If you want to drive this part home, then would be better to test this in `test_pollFunds_fundsEachPartAcrossSchedule` by calling `pollFunds` twice

src/types/ComposableCowPoller.sol
kaze-cow: If the user creates a new TWAP schedule without changing any of hte other parameters (salt, owner, funder, and handler), then they are creating a new TWAP schedule, and technically the old TWAP schedule, while still active, would not be able to receive new funds

Since there is a new TWAP schedule, that would mean that this new TWAP schedule would have its own funds pulling schedule, so it seems reasonable to me that 

At this point, I think its OK to leave the existing functionality, and instead advise the user that `salt` shouldn't be an arbitrary value only used to prevent duplication; it should be the hash of the `staticInput`, with any `appData` related fields set to 0. This more or less completely prevents the collisions we are concerned about.

kaze-cow (changes requested):
im getting a lot of compilation warnings when I attempt to run `forge build` here in `ComposableCowPoller.sol`

I ran coverage tests and happy to confirm the ComposableCowPoller is fully covered

```
|------------------------------------------------------+------------------+------------------+----------------+-----------------|
| src/types/ComposableCowPoller.sol                    | 100.00% (21/21)  | 100.00% (24/24)  | 100.00% (4/4)  | 100.00% (4/4)   |
|------------------------------------------------------+------------------+------------------+----------------+-----------------|
```

Finally, as this is the last PR, I will confirm, I wasn't able to find any significant security issue with *not* including the `staticInput` in the schedule hashing; worst comes to worst, the tokens are sent from the user's wallet to their CoWShed unexpectedly. the `funder` has to set the schedule for any funds that can leave their wallet, so it seems safe. I would, however, recommend we augment `salt` to be renamed something like `staticInputNormalizedHash` or so as this would mostly prevent any possible issues with TWAP schedules or similar being refreshed, and the poller overwriting instead of creating a new schedule accordingly.

src/types/ComposableCowPoller.sol
igorroncevic: While I prefer this style of reverts, this is not a standard adopted in any CoW repo IIRC.

src/types/ComposableCowPoller.sol
igorroncevic: This is useful only if we have callers that need the result. I’d keep the current interface until that's the case.

src/types/ComposableCowPoller.sol
igorroncevic: `ComposableCoW.domainSeparator()` already is the settlement domain separator (see [here](https://github.com/cowprotocol/composable-cow/blob/main/src/ComposableCoW.sol#L72))

src/types/ComposableCowPoller.sol
igorroncevic: Good point, this is something I've also thought about and I've introduced a fix for this with the new mapping:
`mapping(bytes32 id => mapping(bytes32 digest => bool isFunded)) funded`

what this would allow is to keep track of older legs and if they had been funded, preventing any A -> B -> A (older legs) being replayed.

anxolin: Yep, that was silly! I should track all funded orders. Fully disclousure, I just focused on the PoC and then I broke the prototype on PRs, somhow this escaped, good catch!

src/types/ComposableCowPoller.sol
anxolin: Done in 36c73c7

src/types/ComposableCowPoller.sol
anxolin: Fixed by 36c73c71670d80fc3b3e8ae6757d53839d8eda86

test/ComposableCowPoller.t.sol
anxolin: Yes, doesn't cover that much, I see @igorroncevic deleted and im fine with it. 
I think there's value on testing the funds are polled when balance on shed is 0 and when there's enough balance too (as a sepatate case). 

They are technically 2 scenarios, and gives info to see one passes but not the other. Anyways, as mentioned happy to drop.

test/ComposableCowPoller.t.sol
anxolin: @igorroncevic fixed in test_pollFunds_idempotentWithinPartAfterSettlement (36c73c71670d80fc3b3e8ae6757d53839d8eda86)

anxolin:
@kaze-cow this was a great review, thanks, I consider it a gift 🎁
@igorroncevic thanks so much for fixing the issues pointed out by Kaze. I really appreciated the help here to progress this during my time off, nitpicking, when changing others PR is best to make a PR on top that can be later reviewed and merge on top. Alternativelly, if you really commit in the branch (which generally is bad idea), is best you
- Don't solve multiple issues in a single commit (so is easier to review later and also is simpler for you to close conversations by adding the relevant commit)
- Close conversations by adding the relevant commit or PR (took me a bit to understand the state of things, cause comments were on old code, and I needed to make sure they were addressed).

Anyways, thanks a lot to both for the thorough review

src/types/ComposableCowPoller.sol
```diff
@@ -116,7 +119,7 @@ contract ComposableCowPoller {
         if (digest == lastFunded[id]) return;
         lastFunded[id] = digest;
 
-        order.sellToken.transferFrom(schedule.funder, schedule.owner, order.sellAmount);
+        order.sellToken.safeTransferFrom(schedule.funder, schedule.owner, order.sellAmount);
```
anxolin, line 135: Thanks!

src/types/ComposableCowPoller.sol
anxolin: Agree with all this, except the `hash of the staticInput, with any appData related fields set to 0` why is that? Note the `salt` is the salt from composable cow and is used to link it to the composable cow order (doesn't belong to the poller).

Two TWAPs with identical `staticInput` and identical `salt` don't merely collide in the poller, they are the same TWAP (composalble cow identity is those 2 and the twap contract).

FE already today need to generate this salt and make sure they don't reuse it. Is not simplest to just use some random salt  or generate one from current time?

anxolin: To give early feedback to an UI registering the schedule, we could add this check, cost is neglectible because the register function already writes in that storage, so reading is cheap https://github.com/cowprotocol/composable-cow/pull/132

src/types/ComposableCowPoller.sol
```diff
@@ -71,4 +92,34 @@ contract ComposableCowPoller {
...
+                handler: schedule.handler, salt: schedule.salt, staticInput: schedule.staticInput
+            })
+        );
+
+        // The order must still be authorised; `remove` disables the poller.
+        if (!COMPOSABLE_COW.singleOrders(schedule.owner, ctx)) revert OrderNotLive();
+
+        // The handler yields the current order and reverts outside its window.
+        GPv2Order.Data memory order =
+            schedule.handler.getTradeableOrder(schedule.owner, address(this), ctx, schedule.staticInput, bytes(""));
```
igorroncevic, line 128: Should we apply `_guardCheck` like `ComposableCow` does? 
https://github.com/cowprotocol/composable-cow/blob/36c73c71670d80fc3b3e8ae6757d53839d8eda86/src/ComposableCoW.sol#L241-L248

src/types/ComposableCowPoller.sol
```diff
@@ -71,4 +92,34 @@ contract ComposableCowPoller {
...
+
+        // Re-derive `ctx` on-chain, so `pollFunds(id)` stays independent of the order's `appData`.
+        bytes32 ctx = COMPOSABLE_COW.hash(
+            IConditionalOrder.ConditionalOrderParams({
+                handler: schedule.handler, salt: schedule.salt, staticInput: schedule.staticInput
+            })
+        );
+
+        // The order must still be authorised; `remove` disables the poller.
+        if (!COMPOSABLE_COW.singleOrders(schedule.owner, ctx)) revert OrderNotLive();
```
igorroncevic, line 124: Merkle orders are technically unsupported, is this something we'd want to handle in this version?

src/types/ComposableCowPoller.sol
```diff
@@ -27,14 +36,24 @@ contract ComposableCowPoller {
...
     /// @notice Emitted when a schedule is registered or updated.
     /// @param id The deterministic key of the schedule.
     /// @param owner The conditional-order owner and pull destination.
     /// @param funder The token source that registered the schedule.
     event ScheduleRegistered(bytes32 indexed id, address indexed owner, address indexed funder);
+    event Pulled(bytes32 indexed id, bytes32 orderDigest, uint256 amount);
+
+    constructor(ComposableCoW _composableCow) {
+        COMPOSABLE_COW = _composableCow;
+    }
```
igorroncevic, line 67: Let's validate zero address here too, just in case

src/types/ComposableCowPoller.sol
igorroncevic: On second though, adding a boolean return value is a small interface change and it could be useful down the line. I'm inclined to add it.

src/types/ComposableCowPoller.sol
```diff
@@ -27,14 +36,24 @@ contract ComposableCowPoller {
     /// @dev Keyed by `id == scheduleId(schedule)`, which excludes the order's `appData`.
     mapping(bytes32 => Schedule) public schedules;
 
+    /// @dev `id => digest => funded`. History survives schedule updates so an old order cannot be replayed.
+    mapping(bytes32 => mapping(bytes32 => bool)) public funded;
+
     /// @notice Thrown when someone other than the schedule funder registers or updates a schedule.
     error OnlyFunder();
+    error NoSchedule();
+    error OrderNotLive();
```
igorroncevic, line 51: nit: missing natspec

src/types/ComposableCowPoller.sol
```diff
@@ -27,14 +36,24 @@ contract ComposableCowPoller {
...
 
     /// @notice Emitted when a schedule is registered or updated.
     /// @param id The deterministic key of the schedule.
     /// @param owner The conditional-order owner and pull destination.
     /// @param funder The token source that registered the schedule.
     event ScheduleRegistered(bytes32 indexed id, address indexed owner, address indexed funder);
+    event Pulled(bytes32 indexed id, bytes32 orderDigest, uint256 amount);
```
igorroncevic, line 63: Should we index `orderDigest` too, as we could index that specific leg being pulled?

src/types/ComposableCowPoller.sol
anxolin: I agree with the YAGNI, so i was in the fence, but I decide to make a PR proposal, not cause of the tests, but  because is 77gas and allows us to do conditional contracts that chain an action after the poll if we need; https://github.com/cowprotocol/composable-cow/pull/133

Not strongly oppinionated though on including or not including this PR

src/types/ComposableCowPoller.sol
anxolin: I consider out of scope, but actually would be nice to add to Linear this and explore how to add compatibility in an independent follow up PR if we think is worth adding

src/types/ComposableCowPoller.sol
anxolin: I think we don't ususally do this, for example https://github.com/cowprotocol/contracts/blob/main/src/contracts/GPv2Settlement.sol#L73, the contract won't allow to pull funds to anyone. This affects only deploying time, so I would personally skip

src/types/ComposableCowPoller.sol
anxolin: In this repo we are limited by compiler version. We need v0.8.26 and we are in v0.8.19.
Changing compiler, changes the bytecode, and therefore the composable cow address. 

If one day we re-deploy composable cow, we will modernise this repo

src/types/ComposableCowPoller.sol
anxolin: https://linear.app/cowswap/issue/SC-279/poller-consider-integration-with-merkle-proof-flow

src/types/ComposableCowPoller.sol
anxolin: same as https://github.com/cowprotocol/composable-cow/pull/125#discussion_r3682658272

src/types/ComposableCowPoller.sol
anxolin: yes, gas is acceptable, did the change 87bf7c90e933c95a55538a30de2deea35e348d6d

src/types/ComposableCowPoller.sol
anxolin: Added c84ba5d96cbb97c59f995697e6d2e4bee8aebfcc

src/types/ComposableCowPoller.sol
anxolin: This is a great suggestion, but I don't think we should do it.

On top of taking gas for somethign we don't use (minor), the poller doesn't settle nor reason about isValidSignature. It just pulls for a yielded order.

The poll happens before the attempt to settle, so it is still possible for example a hook or wrapper still "enables these guards", so we would be shooting ourselves in the foor by not allowing to poll.

Also, the poller calls `getTradeableOrder` with empty `offchainInput`, and this might be different of what gets passed later in the actual settlement, again preventing to poll for a valid order

src/types/ComposableCowPoller.sol
igorroncevic: Not the best standard I must say, but definitely not blocking.

igorroncevic (dismissed):

kaze-cow (dismissed):

anxolin:
This PR is merged with 2 approvals, even though it doesnt show. Apparently because of the settings in this repo, when I merge and squashed a dependent PR it dismissed the 2 approvals:
```
{at: 19:32:32, was: "approved", reason: "The base branch was changed.", by: igorroncevic}
{at: 19:32:32, was: "approved", reason: "The base branch was changed.", by: kaze-cow}
S
```
However, the commits are the exact ones that were approved, so merged as it is.

PR #126, opened 2026-07-10
chore: deploy scripts for ComposableCowPoller

anxolin:
Part of #121. Follows on #125.

Deploys the ComposableCowPoller contract. 

Adds the scripts, instructions in the readme and deploys the contract (adding the address in the README with the rest of deployed contracts).

It is also deployed as part of the full prod / anvil stacks.

## Test plan
Try to deploy:

```bash
COMPOSABLE_COW=0xfdaFc9d1902f4e0b84f65F49f244b32b31013b74 \
  forge script script/deploy_ComposableCowPoller.s.sol:DeployComposableCowPoller --rpc-url $ETH_RPC_URL --broadcast -vvvv --verify
```

It should deploy and verify.

For now I deployed only in Gnosis Chain, see https://gnosisscan.io/address/0x5eda08425781c2c39a28faaf963c79487dc91bb1

## Out of scope
Deploy in all networks. I deployed in Gnosis Chain to test it, but otherwise I don't plan to deploy in all, besides there's missing features, so we probably want to wait a bit to deploy in all. Maybe even wait for audit

## Approvals
Something changed in github that auto-dismisses the approvals when doing a rebase even thought there's no changes after the review

image

PR #128, opened 2026-07-23
feat(poller): authorize funder CowShed

igorroncevic:
## What and why

Allow a funder’s canonical CowShed to register and revoke Poller schedules. This enables gasless JIT TWAP setup and cancellation while keeping direct funder calls working.

Related: #125

## Test

`forge test --match-contract ComposableCowPollerTest`

### DIFF
--- a/src/types/ComposableCowPoller.sol
+++ b/src/types/ComposableCowPoller.sol
@@ -7,20 +7,25 @@ import {GPv2SafeERC20} from "cowprotocol/contracts/libraries/GPv2SafeERC20.sol";
 import {ComposableCoW} from "src/ComposableCoW.sol";
 import {IConditionalOrder, IConditionalOrderGenerator} from "src/interfaces/IConditionalOrder.sol";
 
+interface ICowShedFactory {
+    function proxyOf(address owner) external view returns (address);
+}
+
 /// @title ComposableCowPoller - Just-in-time funding for composable conditional orders.
 contract ComposableCowPoller {
     using GPv2SafeERC20 for IERC20;
 
     /// @dev `ComposableCoW` stores the settlement domain separator supplied at deployment.
     ComposableCoW public immutable COMPOSABLE_COW;
+    ICowShedFactory public immutable COW_SHED_FACTORY;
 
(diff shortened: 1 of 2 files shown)


anxolin:
Heads up: I restacked the poller chain after #122 merged to `main`, which included a force-push to `feat/poller-cowshed-auth`. Your two commits are preserved verbatim (`3c36195`/`fde4bbf` → `d96670c`/`9ce7395`), just replayed onto the rebased `feat/poller-topup`.

Also worth noting: #126 was stacked on `3c36195` (your *first* commit) rather than this branch's head, so it was effectively re-adding the constructor code-length checks that `fde4bbf` removed. The restack puts #126 on top of this branch's head, so that's resolved.

One small leftover from the refactor: `test/ComposableCowPoller.t.sol:8` now has an unused `ComposableCoW` import, since the constructor revert tests were removed. `forge lint` flags it.

anxolin (commented):
But this is coupling the poller with cow-shed. It was not a dependency before (was kind of general porpouse).

This excludes other AA approaches, and even other shed implementations. I think we would want to support sheds with different salts, trusted executor, or initalization call. 

Maybe not a huge issue, but if they manage to compromise the shed (changing the trusted executor or something) they can exploit the unlimited approval from this contract and steal users funds. So it changes the trust assumption of your approval. 

What I had in mind was to provide a simple:
`registerWithSignature(Schedule, nonce, deadline, signature)`

Isn't that simpler?

igorroncevic:
> But this is coupling the poller with cow-shed. It was not a dependency before (was kind of general porpouse).

> This excludes other AA approaches, and even other shed implementations. I think we would want to support sheds with different salts, trusted executor, or initalization call.

> What I had in mind was to provide a simple:
> registerWithSignature(Schedule, nonce, deadline, signature)

> Isn't that simpler?

Good point! I wanted to keep it way simpler, but it was definitely not inclusive of any alternative approaches and we definitely don't need any additional coupling. I'll close this PR and open a set of 2 new ones for registering/revoking with signature. Thanks!

PR #130, opened 2026-07-29
feat(poller): log the order key on register

anxolin:
A small variation on #129. Alternative to address https://github.com/cowprotocol/composable-cow/pull/123#discussion_r3600738703

Rather than `handler`, `salt` and `keccak256(staticInput)`, it logs `ctx`, the ComposableCoW order key:

```solidity
event ScheduleRegistered(
    bytes32 indexed id, 
    address indexed owner, 
    address indexed funder, 
    bytes32 paramsHash // new
);
```

`paramsHash` is what `singleOrders`, `cabinet` and `remove` are keyed by, identifies the composable cow order.

Composable cow already emits the params ( see https://github.com/cowprotocol/composable-cow/blob/main/src/ComposableCoW.sol#L119) so an indexer could have the full picture with the 2 events.

Gas: **`register` +2,246**, `pollFunds` unchanged. Given the small difference I kind of prefer this approach to https://github.com/cowprotocol/composable-cow/pull/129

## Test plan

```bash

# see the cost: compare the `register` row on the base and on this branch
git checkout feat/poller-ctx-hash           && forge test --gas-report --match-path 'test/ComposableCowPoller.t.sol'
git checkout feat/poller-register-event-ctx && forge test --gas-report --match-path 'test/ComposableCowPoller.t.sol'
```

| | base | this PR | delta |
|---|---|---|---|
| `register` median | 303,215 | 305,461 | **+2,246** |
| `pollFunds` median | 122,132 | 122,132 | **0** |

## NOTE: Approvals

This PR also removed the approvals the moment I rebased the PR (without changes).

image

### DIFF
--- a/src/types/ComposableCowPoller.sol
+++ b/src/types/ComposableCowPoller.sol
@@ -54,7 +54,11 @@ contract ComposableCowPoller {
     /// @param id The deterministic key of the schedule.
     /// @param owner The conditional-order owner and pull destination.
     /// @param funder The token source that registered the schedule.
-    event ScheduleRegistered(bytes32 indexed id, address indexed owner, address indexed funder);
+    /// @param paramsHash The ComposableCoW order key this schedule funds. `id` deliberately excludes
+    ///        `staticInput`, so re-registering the same funder, handler, owner, and salt replaces
+    ///        the stored schedule. Logging the hash names the order each registration points at,
+    ///        which is what makes such a replacement visible off-chain.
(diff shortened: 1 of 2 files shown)


igorroncevic (dismissed):

kaze-cow (dismissed):
as I commented in #130 PR it would be great if we renamed `ctx` as its confusing what it actually represents and typically variables named `ctx` are used for the particular use of providing useful information that needs to get passed around places anyway.

PR #131, opened 2026-07-29
perf(poller): derive the order key locally

anxolin:
Applies a small gas improvement on the way we get the `ctx` on the polling. Polling is done multiple times, so any cost saving is appreciated :) 

## NOTE: Regarding approvals
This PR lost the approvals when rebasing to main after squashing the dependant PR. 
It had one approval from Igor, and a nitpick from Kaze. I will merge, because 2 PRs depend on this and I would like to consolidate these PRs asap. It is a minor change. 

```
21:06:02  review_dismissed        reason: "The base branch was changed."   was: approved
```

## Context
`pollFunds` got `ctx` by calling `ComposableCoW.hash(params)`. That is an external call whose arguments must be ABI-encoded as calldata, and `staticInput` is dynamic, so the encoding dominates the cost.

`ComposableCoW.hash` is `public pure` and just returns `keccak256(abi.encode(params))`, so this mirrors it in a `_ctx` helper. The value is identical — only the call is avoided. If it ever diverged, `pollFunds` would stop finding the order in `singleOrders` and revert `OrderNotLive`, so every polling test already checks the equivalence.

| | before | after | delta |
|---|---|---|---|
| `pollFunds` | 124,640 | 122,132 | **−2,508** |
| bytecode size | 6,147 | 6,047 | **−100 bytes** |

`register` is untouched. Cheaper and smaller, so worth doing on its own. Split out of #130 so the event-field choice there can be judged on its own cost.

## Test plan

```bash

# see the saving: snapshot the base, then diff this branch against it
git checkout feat/poller-deploy   && forge snapshot --snap /tmp/base.gas --match-path 'test/ComposableCowPoller.t.sol'
git checkout feat/poller-ctx-hash && forge snapshot --diff /tmp/base.gas --match-path 'test/ComposableCowPoller.t.sol'
```

It decreases gas for 7 of the 8 tests that poll funds — `RevertWhen_noSchedule` is flat because it reverts before the changed line. Everything that does not poll is flat too, which is what shows `register` is untouched.

image

### DIFF
--- a/src/types/ComposableCowPoller.sol
+++ b/src/types/ComposableCowPoller.sol
@@ -106,6 +106,24 @@ contract ComposableCowPoller {
         delete schedules[id];
     }
 
+    /// @notice Derives the ComposableCoW order key a schedule funds.
+    /// @dev Mirrors `ComposableCoW.hash(params)` rather than calling it: the value is identical,
+    ///      and deriving it here avoids re-encoding `staticInput` as calldata for an external
+    ///      call. The equivalence is exercised by the `pollFunds` tests, which revert
+    ///      `OrderNotLive` if the derived key does not match the authorised order.
+    /// @return The order key, as used by `singleOrders`, `cabinet`, and `remove`.
+    function _ctx(IConditionalOrderGenerator handler, bytes32 salt, bytes memory staticInput)
+        internal
+        pure
+        returns (bytes32)
+    {
+        return keccak256(
+            abi.encode(
+                IConditionalOrder.ConditionalOrderParams({handler: handler, salt: salt, staticInput: staticInput})
+            )
+        );
+    }
+
     /// @notice Move the current order's `sellAmount` from the funder to the owner. Permissionless.
     ///         The full amount always moves (no balance check), so one owner can serve several
     ///         concurrent orders.
@@ -114,11 +132,7 @@ contract ComposableCowPoller {
         if (schedule.funder == address(0)) revert NoSchedule();
 
         // Re-derive `ctx` on-chain, so `pollFunds(id)` stays independent of the order's `appData`.
-        bytes32 ctx = COMPOSABLE_COW.hash(
-            IConditionalOrder.ConditionalOrderParams({
-                handler: schedule.handler, salt: schedule.salt, staticInput: schedule.staticInput
-            })
-        );
+        bytes32 ctx = _ctx(schedule.handler, schedule.salt, schedule.staticInput);
 
         // The order must still be authorised; `remove` disables the poller.
         if (!COMPOSABLE_COW.singleOrders(schedule.owner, ctx)) revert OrderNotLive();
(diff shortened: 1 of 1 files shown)


igorroncevic (dismissed):
LGTM

src/types/ComposableCowPoller.sol
```diff
@@ -123,6 +123,24 @@ contract ComposableCowPoller {
         return funder != address(0) && (msg.sender == funder || msg.sender == COW_SHED_FACTORY.proxyOf(funder));
     }
 
+    /// @notice Derives the ComposableCoW order key a schedule funds.
+    /// @dev Mirrors `ComposableCoW.hash(params)` rather than calling it: the value is identical,
+    ///      and deriving it here avoids re-encoding `staticInput` as calldata for an external
+    ///      call. The equivalence is exercised by the `pollFunds` tests, which revert
+    ///      `OrderNotLive` if the derived key does not match the authorised order.
+    /// @return The order key, as used by `singleOrders`, `cabinet`, and `remove`.
+    function _ctx(IConditionalOrderGenerator handler, bytes32 salt, bytes memory staticInput)
```
kaze-cow: the original function on composable cow was called `hash`, so it makes sense that this function which is performing a similar type of operation should be named similarly. I know that this function ultimately populates a value titled `bytes32 ctx`, but tbh I think the name does not correspond to what most people would correspond to a "context" in a technical sense.

src/types/ComposableCowPoller.sol
```diff
@@ -131,11 +149,7 @@ contract ComposableCowPoller {
         if (schedule.funder == address(0)) revert NoSchedule();
 
         // Re-derive `ctx` on-chain, so `pollFunds(id)` stays independent of the order's `appData`.
-        bytes32 ctx = COMPOSABLE_COW.hash(
-            IConditionalOrder.ConditionalOrderParams({
-                handler: schedule.handler, salt: schedule.salt, staticInput: schedule.staticInput
-            })
-        );
+        bytes32 ctx = _ctx(schedule.handler, schedule.salt, schedule.staticInput);
```
kaze-cow: We should rename this to `paramsHash` or so

kaze-cow (changes requested):

src/types/ComposableCowPoller.sol
anxolin: 949956e70f872e9451c19ee9feb23953eab24227

src/types/ComposableCowPoller.sol
anxolin: 949956e70f872e9451c19ee9feb23953eab24227

src/types/ComposableCowPoller.sol
```diff
@@ -36,7 +36,7 @@ contract ComposableCowPoller {
     /// @dev Keyed by `id == scheduleId(schedule)`, which excludes the order's `appData`.
     mapping(bytes32 => Schedule) public schedules;
 
-    /// @dev `id => digest => funded`. History survives schedule updates so an old order cannot be replayed.
```
anxolin, line 39: sorry for this unrelated change. But since im changing names to make it more readable, I changed this one that bother me too

PR #132, opened 2026-07-29
feat(poller): reject registering a taken key

anxolin:
Follow up to #125, addressing https://github.com/cowprotocol/composable-cow/pull/125#discussion_r3578723006 — Igor's suggestion to reject overriding existing IDs.

The schedule key ignores `staticInput`, so registering over an existing key replaced the stored schedule and left the previous, still authorised, order unfundable. Nothing in the logs showed it: that `ScheduleRegistered` looked like any other.

Now it reverts with `AlreadyRegistered`. To replace a schedule deliberately, `revoke` first — that emits `ScheduleRevoked` and can be batched with the new `register` in one transaction. `funded[id][digest]` is not cleared by `revoke`, so replay protection survives the cycle.

Gas: **`register` +201** (0.07%), +95 bytes. Nearly free, since `register` already writes that slot.

Also rewrites the `salt` natspec, which called it "a user-controlled namespace" and recommended deriving it from the static input with appData zeroed. It is really the conditional order's own `salt` — the same value ComposableCoW hashes into `ctx` — so that derivation would give two identical orders the same key and stop them coexisting, and the contract can never verify it anyway. It now states the constraint and recommends a fresh random value per order. Discussion: https://github.com/cowprotocol/composable-cow/pull/125#issuecomment-5123134104

## Test plan

```bash
forge test --match-path 'test/ComposableCowPoller.t.sol'   # 24 passed
```

`test_register_replacesScheduleWithSameId` becomes `test_register_RevertWhen_alreadyRegistered`, plus `test_register_afterRevokeReusesTheKey` for the deliberate-replacement path.

## NOTE: Approvals

Again, it just removed Igors approval when I merged the dependant PR

image

### DIFF
--- a/src/types/ComposableCowPoller.sol
+++ b/src/types/ComposableCowPoller.sol
@@ -15,7 +15,9 @@ contract ComposableCowPoller {
     ComposableCoW public immutable COMPOSABLE_COW;
 
     /// @notice Parameters for a JIT funding schedule.
-    /// @dev A schedule is uniquely identified by its funder, handler, owner, and salt.
+    /// @dev `handler`, `salt` and `staticInput` are the order's `ConditionalOrderParams` and must
+    ///      match it exactly, since `paramsHash` is derived from them. The schedule key is `funder`,
+    ///      `handler`, `owner`, and `salt`.
     struct Schedule {
         /// @notice The conditional-order handler to poll, such as the TWAP type.
         IConditionalOrderGenerator handler;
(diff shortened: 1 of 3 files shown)


src/types/ComposableCowPoller.sol
```diff
@@ -30,11 +32,12 @@ contract ComposableCowPoller {
         /// @notice The address that owns the ComposableCoW conditional order and receives the pulled funds.
         /// @dev It can be an EOA or contract and may be the same address as `funder`.
         address owner;
-        /// @notice A user-controlled namespace for this schedule.
-        /// @dev Keep it unique to the user and logical order. A good default is the hash of the
-        ///      order-defining static-input values with any appData field set to zero.
+        /// @notice The conditional order's own `salt`.
+        /// @dev It is what keeps two otherwise-identical orders distinct in ComposableCoW, so use
+        ///      a fresh random value per order.
         bytes32 salt;
```
anxolin, line 33: note I  tried to explain where `salt` `handler`, and `staticInput` come from, and removed the recomendation of hashing the static input

igorroncevic: We should also update README.md to remove this recommendation, and note that salt should be unique there too.

anxolin: https://github.com/cowprotocol/composable-cow/pull/132/changes/9415b16885820aa7d2431737b9b7c09e6372f9c6

igorroncevic (dismissed):

PR #133, opened 2026-07-30
feat(poller): finalise the pollFunds observable surface

anxolin:
Follow up to #125, addressing https://github.com/cowprotocol/composable-cow/pull/125#discussion_r3601055050 

`pollFunds` has two successful outcomes: funds moved, or this order was already funded and nothing happened. 

An on-chain caller cannot tell them apart without diffing balances before and after. 

Every other case reverts, so a bool says it exactly.

I was in the fence with doing or not doing, because we don't need it right now (for TWAP for EOAs, since this is a direct call in the hooks), but:
- It adds just 77 gas units, insignificant
- Allows future use cases where we wrap this contract to do some action after the funds have been moved. 

| | before | after | delta |
|---|---|---|---|
| `pollFunds` median | 122,132 | 122,209 | **+77** |
| bytecode | 6,047 | 6,060 | +13 bytes |

`register` is unaffected.

## Also: index `orderDigest` on `Pulled`

Addressing https://github.com/cowprotocol/composable-cow/pull/125#discussion_r3681587079 — same reasoning as above, an ABI decision that can only be made before deployment.

Only `id` was indexed, so every query had to start from the schedule. `orderDigest` is the CoW order struct hash, so indexing it lets a single leg be looked up directly and joined to settlement data. `indexed` does not change the event signature, so a consumer built against the old layout would silently misdecode rather than fail.

| | before | after | delta |
|---|---|---|---|
| `pollFunds` median | 122,209 | 122,307 | **+98** |
| bytecode | 6,060 | 6,052 | **−8 bytes** |

Smaller, because the value moves out of the ABI-encoded data.

`Pulled` had no test at all, so its layout was unverified. `test_pollFunds_movesFullAmountUnconditionally` now asserts it — I checked the assertion fails on both a wrong digest and a wrong amount.

Combined effect of both commits versus #131: `pollFunds` 122,132 → 122,307 (+175), bytecode 6,047 → 6,052 (+5).

## Test plan

```bash
```

## NOTE: Approvals

This is annoying, not sure which setting is doing this, but  every time I merge a dependant PR I loose the approvals without changing any code in the branch.  

image

### DIFF
--- a/src/types/ComposableCowPoller.sol
+++ b/src/types/ComposableCowPoller.sol
@@ -67,11 +67,14 @@ contract ComposableCowPoller {
     ///        which is what makes such a replacement visible off-chain.
     event ScheduleRegistered(bytes32 indexed id, address indexed owner, address indexed funder, bytes32 paramsHash);
 
-    /// @notice Emitted when an order's sell amount is moved from the funder to the owner.
-    /// @param id The deterministic key of the schedule that was polled.
-    /// @param orderDigest The EIP-712 digest of the funded order, unique per part of the schedule.
-    /// @param amount The amount of `sellToken` transferred.
-    event Pulled(bytes32 indexed id, bytes32 orderDigest, uint256 amount);
+    /// @notice Emitted when a part's funds are moved.
+    /// @dev `orderDigest` is indexed so a single leg can be looked up directly, rather than only
+    ///      by walking a schedule's history. It is the CoW order struct hash, so it joins to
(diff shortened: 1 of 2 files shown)


igorroncevic (dismissed):
LGTM, I like the distinction between non-existing schedule (bad state) and existing one that has already been pulled (good state).

kaze-cow (dismissed):

PR #135, opened 2026-08-03
feat(poller): add signed schedule registration

igorroncevic:
## What and why

Funders can authorize any relayer to register a complete polling schedule with an EIP-712 signature. Registrations support both EOAs and ERC-1271 smart accounts.

The signed schedule includes its `authEpoch`. Revocation increments the stored epoch, invalidating authorizations from prior epochs while allowing the same schedule ID to be reused. The typed-data schema exposes `staticInput` as `bytes` and hashes its value during EIP-712 encoding.

Depends on #133.

## Test

```bash
```


src/types/ComposableCowPoller.sol
```diff
@@ -108,18 +124,61 @@ contract ComposableCowPoller {
...
+        if (!SignatureChecker.isValidSignatureNow(schedule.funder, _hashTypedDataV4(structHash), signature)) {
+            revert InvalidSignature();
+        }
+
+        id = scheduleId(schedule);
+        return _register(schedule, id);
+    }
+
+    /// @dev Stores a schedule and consumes one action nonce for its funder.
+    function _register(ComposableCowPoller.Schedule calldata schedule, bytes32 id)
```
anxolin: why receiving the scheule and the id, once thing is derived by the other, and the id is returned anyways plus the only two uses of `_register` both do `id = scheduleId(schedule);`, so I believe the internal method should only have one parameter; the schedule

src/types/ComposableCowPoller.sol
```diff
@@ -108,18 +124,61 @@ contract ComposableCowPoller {
...
+        internal
+        returns (bytes32)
+    {
         if (schedules[id].funder != address(0)) revert AlreadyRegistered();
+        nonces[schedule.funder]++;
```
anxolin: why incrementing the nonces? this is problematic because it prevent us from having more than one in-flight orders. So a user can't place a couple of TWAPs for example and needs to wait for the sell=buy order to execute.

Also, registering conflicts with the revoke, so they compete

src/types/ComposableCowPoller.sol
```diff
@@ -108,18 +124,61 @@ contract ComposableCowPoller {
...
+        if (block.timestamp > deadline) revert SignatureExpired();
+
+        bytes32 structHash = keccak256(
+            abi.encode(
+                REGISTER_TYPEHASH,
+                schedule.handler,
+                schedule.funder,
+                schedule.owner,
+                schedule.salt,
+                keccak256(schedule.staticInput),
```
anxolin, line 159: do you need to hash it or can we pass directly `staticInput`?

anxolin (changes requested):

src/types/ComposableCowPoller.sol
igorroncevic: I think hashing is important because bytes are dynamic so it's best to represent them as a smaller "string" in a form of a hash.

src/types/ComposableCowPoller.sol
igorroncevic: Good point. I don't anticipate much concurrency here but I guess it's not the best UX.

I still think we need some replay protection, so I propose we don't allow IDs to be reused by requiring `schedules[id].funder` to be `address(0)` on registration.

src/types/ComposableCowPoller.sol
```diff
@@ -147,40 +140,38 @@ contract ComposableCowPoller is EIP712 {
...
     }
 
-    /// @dev Stores a schedule and consumes one action nonce for its funder.
-    function _register(ComposableCowPoller.Schedule calldata schedule, bytes32 id)
-        internal
-        returns (bytes32)
-    {
+    /// @dev Stores a schedule and permanently marks its ID as used.
+    function _register(ComposableCowPoller.Schedule calldata schedule) internal returns (bytes32 id) {
+        if (address(schedule.handler) == address(0)) revert InvalidHandler();
```
anxolin: Do we care if someone registers the handler 0? it will never be able to poll for it, no?

src/types/ComposableCowPoller.sol
```diff
@@ -147,40 +140,38 @@ contract ComposableCowPoller is EIP712 {
...
             id, schedule.owner, schedule.funder, _paramsHash(schedule.handler, schedule.salt, schedule.staticInput)
         );
-        return id;
     }
 
     /// @notice Revoke a schedule. Only the funds source may do so. A standing ERC-20 allowance
     ///         should be revoked separately to fully close the surface.
     function revoke(bytes32 id) external {
         Schedule storage schedule = schedules[id];
+        if (address(schedule.handler) == address(0)) revert NoSchedule();
```
anxolin: Do we care if someone deletes a schedule whose handler is 0?

anxolin: Also, asserted in creation

src/types/ComposableCowPoller.sol
```diff
@@ -211,7 +202,7 @@ contract ComposableCowPoller is EIP712 {
     ///         balances; every other case reverts.
     function pollFunds(bytes32 id) external returns (bool) {
         Schedule memory schedule = schedules[id];
-        if (schedule.funder == address(0)) revert NoSchedule();
+        if (address(schedule.handler) == address(0)) revert NoSchedule();
```
anxolin: I didn't get this change, why using then handler? 
The registration already checks the founder is not the 0x00 address

src/types/ComposableCowPoller.sol
anxolin: Yes, I think hashing is needed to encode the data;

Image

But I think the issue, is that, in the typehash we can use `bytes` and a per spec the wallet knows how to encode and hash the data. 

So I think we need to change from `bytes32` to `bytes`, otherwise apps will need to hash it before passing the signing data to the wallet, obfuscating what u actually sign.

Note they make a distintion between encoding and type definition for dynamic types;

Image

src/types/ComposableCowPoller.sol
```diff
@@ -119,9 +165,13 @@ contract ComposableCowPoller {
...
     function revoke(bytes32 id) external {
         Schedule storage schedule = schedules[id];
+        if (address(schedule.handler) == address(0)) revert NoSchedule();
         if (msg.sender != schedule.funder) revert OnlyFunder();
         emit ScheduleRevoked(id, schedule.owner, schedule.funder);
-        delete schedules[id];
+        delete schedule.handler;
+        delete schedule.owner;
+        delete schedule.salt;
+        delete schedule.staticInput;
```
anxolin: I only realised later. But on the check you do about handler not being 0x00, is because u deliverably leave the founder to prevent someone from recreating the schedule twice

I think this is undedocumented in the PR description, but most importantly in the natspec of this fuction, and in the delete section.

Probably nice also to add it in `schedules` definition that `funder` outlives revocation as a tombstone.

src/types/ComposableCowPoller.sol
anxolin: Only thing I come up with here that could be bad, is too nitpic, is to prevent the emission of the event

src/types/ComposableCowPoller.sol
anxolin: closing, but good to paste the commit that address

src/types/ComposableCowPoller.sol
anxolin: Now i get it, but I was confused for a bit until I understood what u did. I was unsure how `schedules[id].funder to be address(0) on registration` help here.

What u did is different, is to leave the schedule `founder` un-deleted, just to u know the schedule was created before. 

Also, u changed the logic. This should be highlighted. I also think this can become a problem. Imagine u create the TWAP with a salt. U create one schedule, but u want to change the schedule, now u cannot delete and re-create it, because to re-create u need a new salt, and that requires a new TWAP.

Previously, https://github.com/cowprotocol/composable-cow/pull/132 deleted the entire registry, which allowed you to delete and recreate it. Off the top of my head, I'm not opposed to the new behaviour, but this should be highlighted.

anxolin: Also, the description of the PR still describes the `nonce` behaviour. The PR description becomes part of the documentation that explains the history of the code, so is good is in sync with the changes

anxolin (changes requested):

src/types/ComposableCowPoller.sol
igorroncevic: Good point! I'll make it into `bytes` instead of `bytes32` so users sign exactly the unobfuscated field.

src/types/ComposableCowPoller.sol
igorroncevic: Agreed. The shared nonce has been removed, so independent schedule registrations no longer invalidate each other.

Replay protection is now per ID:

- `funder != address(0)` means the ID has been permanently used (`funder` is kept when revoking)
- `handler != address(0)` means the schedule is currently active.

This intentionally prevents an old registration signature from restoring a revoked schedule. The tradeoff is that recreating or changing the TWAP requires a new salt and therefore a new order.

I’m addressing the pre-registration revoke in #136 with revoke that works whether registration lands before or after cancellation. I’ll also update this PR’s description to remove the stale nonce behavior.

src/types/ComposableCowPoller.sol
igorroncevic: See https://github.com/cowprotocol/composable-cow/pull/135#discussion_r3765659996

src/types/ComposableCowPoller.sol
igorroncevic: See https://github.com/cowprotocol/composable-cow/pull/135#discussion_r3765659996

src/types/ComposableCowPoller.sol
igorroncevic: See https://github.com/cowprotocol/composable-cow/pull/135#discussion_r3765659996 it's part of the new replay protection mechanism

src/types/ComposableCowPoller.sol
igorroncevic: Will do!

src/types/ComposableCowPoller.sol
igorroncevic: Fixed! Replaced shared nonces with independent single-use schedule IDs and updated the PR description.

https://github.com/cowprotocol/composable-cow/commit/5522dea8d3eec1fe2695f91e3c5ca866ca04e81b

src/types/ComposableCowPoller.sol
igorroncevic: Fixed! The typed-data field is now bytes, while EIP-712 encoding still hashes its value. Added an independent signature test.

https://github.com/cowprotocol/composable-cow/commit/a338f52c1ba6a24d9dc21f5fff91e0c2b12e4b2a

src/types/ComposableCowPoller.sol
igorroncevic: Fixed! Added NatSpec for the funder tombstone on the schedule mapping and revoke function.

https://github.com/cowprotocol/composable-cow/commit/a338f52c1ba6a24d9dc21f5fff91e0c2b12e4b2a

anxolin: nit, i think is nice also to do what I hinted above. To add a comment also in the `delete` sectition, to ratify that `delete schedule.funder;` is omited on porpouse

anxolin (commented):
Soft aaprove, 2 small things u might want to address

(1) There's somethign I can shake off. 
This thing with the thombstomb feels a bit not intuitive and easy to miss. This feels a bit hacky/fragile;
```
funder != address(0) means the ID has been permanently used (funder is kept when revoking)
handler != address(0) means the schedule is currently active.
```

At first I thought we could create a mapping for burned ids, which is simpler and easier to reason about, but this adds a bit of gas compared to this solution. 

Maybe a compromise is to add state to the struct. Like `enum Status { None, Active, Revoked }`. If done right, most of the time doesn't add gas, and only on revocations we consume a bit of extra gas. Revocations won't happen often. 

EDIT: While reviewing #135 I just thought a bit more about the revocation. Is a pitty we changed the implementation to make the revocation final in these PRs just to add the signatures. I think we can kill the thombstone need (making the thing cleaner) and let the user recreate the deleted schedule after deleting, and still handle the "in flight leaked signatures" problem all by adding a ` uint96 generation;` in the `StoredSchedule ` struct right after `handler` (so it still uses the slot 0).  This way, when you sign the 712 messsage u should include the "generation" of the the schedule you used. Same applies for #136 revoke message. Note I mentioned `StoredSchedule ` cause we differenciate now the stored struct from the `StoredSchedule ` used in the register function for example. 

(2) I believe the pr needs to be rebased with `main`, is based on an already merged PR

src/types/ComposableCowPoller.sol
```diff
@@ -3,14 +3,21 @@ pragma solidity >=0.8.0 <0.9.0;
...
 import {IConditionalOrder, IConditionalOrderGenerator} from "src/interfaces/IConditionalOrder.sol";
 
 /// @title ComposableCowPoller - Just-in-time funding for composable conditional orders.
-contract ComposableCowPoller {
+contract ComposableCowPoller is EIP712 {
     using GPv2SafeERC20 for IERC20;
 
+    /// @dev EIP-712 Register struct typehash.
+    bytes32 private constant REGISTER_TYPEHASH = keccak256(
+        "Register(address handler,uint96 authEpoch,address funder,address owner,bytes32 salt,bytes staticInput,uint256 deadline)"
```
kaze-cow: it seems a bit unusual there is no actual structure to go along with this defined EIP-712 data structure. probably fine (I see the abi.decode is happening later), but surprised me.

kaze-cow: i also just noticed this structure is the same as `Schedule`, except without the `deadline` field. So maybe the real name of this typehash should be `ScheduleRegistration` instead of `Register`

src/types/ComposableCowPoller.sol
```diff
@@ -21,6 +28,10 @@ contract ComposableCowPoller {
     struct Schedule {
         /// @notice The conditional-order handler to poll, such as the TWAP type.
         IConditionalOrderGenerator handler;
+        /// @notice The authorization epoch for this schedule ID.
+        /// @dev Starts at zero and must match the stored value. Revocation increments it, allowing
```
kaze-cow: what must match the stored value? the payload on `poll`?

src/types/ComposableCowPoller.sol
```diff
@@ -87,41 +110,78 @@ contract ComposableCowPoller {
...
+    ///      match storage, so revocation invalidates signatures from prior epochs.
+    /// @param schedule The schedule to store.
+    /// @param deadline The last block timestamp at which the signature is valid.
+    /// @param signature The funder's EIP-712 signature.
+    /// @return id The deterministic key of the stored schedule.
+    function registerWithSignature(Schedule calldata schedule, uint256 deadline, bytes calldata signature)
+        external
+        returns (bytes32 id)
+    {
+        if (block.timestamp > deadline) revert SignatureExpired();
```
kaze-cow, line 149:
```suggestion
        require(block.timestamp <= deadline, SignatureExpired());
```

src/types/ComposableCowPoller.sol
```diff
@@ -87,41 +110,78 @@ contract ComposableCowPoller {
...
+                schedule.funder,
+                schedule.owner,
+                schedule.salt,
+                keccak256(schedule.staticInput),
+                deadline
+            )
+        );
+        if (!SignatureChecker.isValidSignatureNow(schedule.funder, _hashTypedDataV4(structHash), signature)) {
+            revert InvalidSignature();
+        }
```
kaze-cow, line 165:
```suggestion
        require(SignatureChecker.isValidSignatureNow(schedule.funder, _hashTypedDataV4(structHash), signature), InvalidSignature());
```

test/ComposableCowPoller.t.sol
```diff
@@ -111,10 +137,50 @@ contract ComposableCowPollerTest is BaseComposableCoWTest {
...
+                schedule.handler,
+                schedule.authEpoch,
+                schedule.funder,
+                schedule.owner,
+                schedule.salt,
+                keccak256(schedule.staticInput),
+                deadline
+            )
+        );
+        return keccak256(abi.encodePacked("\x19\x01", _domainSeparator(chainId, verifyingContract), structHash));
```
kaze-cow: nit the domain separator can probably be pulled from the actual `ComposableCowPoller` as a public function without actually recalculating it here.

test/ComposableCowPoller.t.sol
```diff
@@ -11,9 +12,30 @@ import {ComposableCowPoller} from "src/types/ComposableCowPoller.sol";
...
+}
+
 /// @title ComposableCowPoller unit tests
 /// @notice Exercises registering a schedule for a composable TWAP created via `createWithContext`.
 contract ComposableCowPollerTest is BaseComposableCoWTest {
+    bytes32 constant EIP712_DOMAIN_TYPEHASH =
+        keccak256("EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)");
+    bytes32 constant REGISTER_TYPEHASH = keccak256(
+        "Register(address handler,uint96 authEpoch,address funder,address owner,bytes32 salt,bytes staticInput,uint256 deadline)"
+    );
```
kaze-cow: repeated in the test file. Probably OK to use the same def in the `ComposableCowPoller`

src/types/ComposableCowPoller.sol
```diff
@@ -21,6 +28,10 @@ contract ComposableCowPoller {
     struct Schedule {
         /// @notice The conditional-order handler to poll, such as the TWAP type.
         IConditionalOrderGenerator handler;
+        /// @notice The authorization epoch for this schedule ID.
+        /// @dev Starts at zero and must match the stored value. Revocation increments it, allowing
+        ///      the same ID to be reused while invalidating signatures from prior epochs.
+        uint96 authEpoch;
```
kaze-cow, line 35: this is, in effect, a sort of nonce, right? can we use the industry wide term "nonce" here?

test/ComposableCowPoller.t.sol
```diff
@@ -36,7 +59,9 @@ contract ComposableCowPollerTest is BaseComposableCoWTest {
         twap = new TWAP(composableCow);
         currentBlockTimestampFactory = new CurrentBlockTimestampFactory();
         poller = new ComposableCowPoller(composableCow);
-        funder = makeAddr("funder");
+        funderPrivateKey = uint256(keccak256("funder"));
+        funder = vm.addr(funderPrivateKey);
+        vm.label(funder, "funder");
```
kaze-cow: we have https://www.getfoundry.sh/reference/forge-std/make-addr-and-key

test/ComposableCowPoller.t.sol
kaze-cow, line 1: I might be missing it but I don't see an explicit negative test for what happens when `authEpoch` *doesnt* match up.

test/ComposableCowPoller.t.sol
```diff
@@ -235,7 +315,184 @@ contract ComposableCowPollerTest is BaseComposableCoWTest {
         );
     }
 
-    /// @dev The funder can revoke, which clears the schedule.
+    function test_registerWithSignature_allowsArbitraryCallerWithEOASignature() public {
+        ComposableCowPoller.Schedule memory schedule = _schedule(SALT, abi.encode(_bundle()));
+        uint256 deadline = block.timestamp + 1 hours;
+        bytes memory signature = _signRegister(schedule, deadline);
+
+        vm.prank(bob.addr);
```
kaze-cow: to make it really really clear this is an "arbitrary" caller I would just generate a brand new address here with `vm.makeAddr("arbitrary caller")` or so

test/ComposableCowPoller.t.sol
```diff
@@ -235,7 +315,184 @@ contract ComposableCowPollerTest is BaseComposableCoWTest {
...
+
+    function test_registerWithSignature_RevertWhen_signedForDifferentPoller() public {
+        ComposableCowPoller otherPoller = new ComposableCowPoller(composableCow);
+        ComposableCowPoller.Schedule memory schedule = _schedule(SALT, abi.encode(_bundle()));
+        uint256 deadline = block.timestamp + 1 hours;
+        bytes32 digest = _registerDigest(schedule, deadline, block.chainid, address(otherPoller));
+
+        vm.expectRevert(ComposableCowPoller.InvalidSignature.selector);
+        poller.registerWithSignature(schedule, deadline, _sign(funderPrivateKey, digest));
+    }
```
kaze-cow: you could probably compress this just into different domain selector. Maybe starting to enter the territory of testing openzeppelin's code.

test/ComposableCowPoller.t.sol
```diff
@@ -235,7 +315,184 @@ contract ComposableCowPollerTest is BaseComposableCoWTest {
...
+        vm.expectRevert(ComposableCowPoller.InvalidAuthEpoch.selector);
+        poller.registerWithSignature(schedule, deadline, signature);
+
+        schedule.authEpoch = 1;
+        assertEq(
+            poller.registerWithSignature(schedule, deadline, _signRegister(schedule, deadline)),
+            id,
+            "fresh signature reuses ID"
+        );
+    }
```
kaze-cow, line 450: ah! this was the negative test I was looking for that verifies that `authEpoch` needs to match up/is incremented. Technically this test seems to supersede what is happening earlier in the file in `test_register_afterRevokeReusesTheKey`

kaze-cow (changes requested):
most of my comments are kind of minor. I didn't see any showstopping issues.

src/types/ComposableCowPoller.sol
igorroncevic: It is, but then also it isn't. This value is kept the same for the entire duration of ID's lifecycle until it is explicitly revoked and incremented to allow reuse, when it can be signed again. Sounds like a nonce but to me it's more of an epoch counter for that specific ID.

src/types/ComposableCowPoller.sol
igorroncevic: `Schedule` is the Solidity struct, the signed message is `Schedule` + `deadline` - didn't wanna overcomplicate it. I renamed the EIP-712 type to `ScheduleRegistration`.

src/types/ComposableCowPoller.sol
igorroncevic: Not `pollFunds`. The `Schedule.authEpoch` passed during registration must match the stored epoch for that schedule ID, and I clarified that in the NatSpec.

src/types/ComposableCowPoller.sol
igorroncevic: For all `require` comments - the project uses 0.8.19, while the custom error syntax requires 0.8.26. Not sure if upgrading compiler versions just for this is fine.

src/types/ComposableCowPoller.sol
igorroncevic: https://github.com/cowprotocol/composable-cow/pull/135#discussion_r3805448970

test/ComposableCowPoller.t.sol
igorroncevic: Covered by `test_register_afterRevokeReusesTheKey` and `test_registerWithSignature_RevertWhen_replayedAfterRevoke`.

test/ComposableCowPoller.t.sol
igorroncevic: Done. Removed the earlier test since the signed replay test covers the epoch/reuse lifecycle.

test/ComposableCowPoller.t.sol
igorroncevic: Done. Combined the wrong chain and wrong contract checks into one domain test.

test/ComposableCowPoller.t.sol
igorroncevic: done

test/ComposableCowPoller.t.sol
igorroncevic: done

kaze-cow (approved):

src/types/ComposableCowPoller.sol
```diff
@@ -86,42 +110,84 @@ contract ComposableCowPoller {
...
+    ///      distinct orders on distinct keys.
     /// @param schedule The schedule whose identity fields determine the key.
     /// @return The schedule key.
     function scheduleId(Schedule memory schedule) public pure returns (bytes32) {
         return keccak256(abi.encode(schedule.funder, schedule.handler, schedule.owner, schedule.salt));
     }
 
     /// @notice Registers a schedule.
-    /// @dev Reverts if the key is taken. The key ignores `staticInput`, so silently overwriting
-    ///      would leave the previous — still authorised — order unfundable with nothing in the
```
anxolin, line 100: This still applies, why deleting?

src/types/ComposableCowPoller.sol
```diff
@@ -86,42 +110,84 @@ contract ComposableCowPoller {
...
-    ///      logs to show it. `revoke` first to replace a schedule deliberately; funding history
-    ///      survives that, so an old order cannot be replayed.
-    ///      Only the funds source may register, and the ID is namespaced by the funder.
-    ///      A zero `owner` or `handler` needs no check: `pollFunds` requires
-    ///      `singleOrders[owner][paramsHash]`, which neither can ever satisfy.
```
anxolin, line 105: Also, this is still true, we might want to specify why we don't care about verifying the 0x00 owner?

test/ComposableCowPoller.t.sol
```diff
@@ -235,7 +276,180 @@ contract ComposableCowPollerTest is BaseComposableCoWTest {
...
+        ComposableCowPoller.Schedule memory schedule = _schedule(SALT, abi.encode(_bundle()));
+        uint256 deadline = block.timestamp + 1 hours;
+        bytes memory signature = _signRegister(schedule, deadline);
+        bytes32 id = poller.registerWithSignature(schedule, deadline, signature);
+
+        vm.prank(funder);
+        poller.revoke(id);
+
+        vm.expectRevert(ComposableCowPoller.InvalidAuthEpoch.selector);
+        poller.registerWithSignature(schedule, deadline, signature);
```
anxolin, line 442: Should we add this test also for the "normal register"? Technically it also throws this new error, no?

src/types/ComposableCowPoller.sol
anxolin, line 92: should we add authEpoch here or is best not to?

src/types/ComposableCowPoller.sol
```diff
@@ -86,42 +110,84 @@ contract ComposableCowPoller {
     /// @param funder The token source that revoked the schedule.
     event ScheduleRevoked(bytes32 indexed id, address indexed owner, address indexed funder);
 
+    /// @notice Returns the EIP-712 domain separator used to authorize registrations.
+    function domainSeparator() external view returns (bytes32) {
```
anxolin, line 114: Could we name this `authDomainSeparator` or `pollerDomainSeparator` to differenciate from `COMPOSABLE_COW.domainSeparator(),` also used in this file?

anxolin (approved):
Soft approve. Only minor comments!

src/types/ComposableCowPoller.sol
igorroncevic: Addressed in https://github.com/cowprotocol/composable-cow/commit/3385c621a37f307dfe8bef2a3a5288d33053f23f

src/types/ComposableCowPoller.sol
igorroncevic: Addressed in https://github.com/cowprotocol/composable-cow/commit/782e8994337b09cc3df67252ad12892fb0d6a07d

test/ComposableCowPoller.t.sol
igorroncevic: Addressed in https://github.com/cowprotocol/composable-cow/commit/2883dc2db85ca5e8d213055f262858d829f09866

src/types/ComposableCowPoller.sol
igorroncevic: Addressed in https://github.com/cowprotocol/composable-cow/commit/8eef40008c73ac2e3f7bbc28c3fd8e17020e1914

src/types/ComposableCowPoller.sol
igorroncevic: I don't think that's necessary, `domainSeparator` is a pretty standard function + the two won't collide whatsoever.

PR #136, opened 2026-08-03
feat(poller): add signed schedule revocation

igorroncevic:
## What and why

Funders can revoke polling schedules through relayers with an EIP-712 signature. The authorization binds the schedule ID, funder, and deadline. Schedule IDs are permanently single-use, so a direct or signed revocation invalidates every other revocation signature for that ID without affecting other schedules. Both EOA and ERC-1271 signatures are supported.

Depends on #135.

## Test

```bash
```


src/types/ComposableCowPoller.sol
```diff
@@ -178,6 +182,30 @@ contract ComposableCowPoller is EIP712 {
...
         if (msg.sender != schedule.funder) revert OnlyFunder();
+        _revoke(id, schedule);
+    }
+
+    /// @notice Revokes a schedule authorized by the funder's EIP-712 signature.
+    /// @dev Any caller may submit the signature before its deadline.
+    /// @param id The deterministic key of the schedule to revoke.
+    /// @param deadline The last block timestamp at which the signature is valid.
+    /// @param signature The funder's EIP-712 signature.
+    function revokeWithSignature(bytes32 id, uint256 deadline, bytes calldata signature) external {
```
anxolin: So we sign for revoking too, thats nice, but shouldt' we be able to revoke using a direct transaction too? just to invalidate the nonce

anxolin (changes requested):

src/types/ComposableCowPoller.sol
igorroncevic: but we are able to do that, there's a `revoke(bytes32 id)` right above this one that calls the same `_revoke` function as `revokeWithSignature`

anxolin: But I don't think it behaves as someone would expect, no? 

The revokes currently zeros the struct fields, but imagine the schedule was not created when you revoke. Anyone with a signature for the creation can create it, even if I do the revoke. 

Imagine I send the register schedule to the API. Anyone can see the signature. I need to revoke my signature, so I do an on-chain revoke. Anyone with my signature can just create it after that.

anxolin (changes requested):

src/types/ComposableCowPoller.sol
igorroncevic: Great point! I didn't think about the flow like that, I'll make sure to add a pre-registration revoke path because of this. Thanks!

anxolin: I believe this comment was placed for this conversation, so I wll add here for reference

> @anxolin Added pre-registration revocation for both direct and signed paths. The schedule ID is derived from the authenticated funder, handler, owner, and salt, and the retained funder acts as the used-ID tombstone, so pending registration signatures cannot land afterward.

> Could you please take another look?
> https://github.com/cowprotocol/composable-cow/commit/6b056ef76738d87482f518c55ee0b365441a7007

anxolin (commented):
There was some things it felt off with all this thombstone bussiness, I think a lot of it goes away if we do (1) of this comment from #135  https://github.com/cowprotocol/composable-cow/pull/135#pullrequestreview-4929976430

Whatever we decided in #135 will affect this PR. 

I think we are almost there!

src/types/ComposableCowPoller.sol
igorroncevic: Yup, thanks

src/types/ComposableCowPoller.sol
```diff
@@ -178,16 +182,56 @@ contract ComposableCowPoller is EIP712 {
...
+    /// @param deadline The last block timestamp at which the signature is valid.
+    /// @param signature The funder's EIP-712 signature.
+    function revokeWithSignature(
+        IConditionalOrderGenerator handler,
+        address funder,
+        address owner,
+        bytes32 salt,
+        uint256 deadline,
+        bytes calldata signature
+    ) external returns (bytes32 id) {
```
anxolin, line 219: Should this function take also the `authEpoch`, otherwise we can't tell the difference between invalid signature and passing the wrong authEpoch, which can happen if for example is bumped

anxolin (approved):
Approved, but addeda small suggestion

src/types/ComposableCowPoller.sol
igorroncevic: Addressed in https://github.com/cowprotocol/composable-cow/commit/ba94ef526a76ff4009d7825de6d1c204e46a57f8

PR #103, opened 2025-07-24
Lens deployment

squadgazzz:
# Description
This PR contains the Lens chain deployment info. No broadcast data since SCs were deployed and verified manually.

PR #139, opened 2026-08-06
feat: generate networks.json from the deployment artifacts

anxolin:
Closes the loop on https://github.com/cowprotocol/composable-cow/pull/103#issuecomment-3161290115.

`networks.json` has been maintained by hand since the generator script was deleted in #103. It was deleted because the Lens contracts were deployed manually and left no broadcast artifacts for the script to read, so the file had to be edited directly.

This PR restores the script

## Changes

- `broadcast/networks-manual.json`: 
   - chains deployed by hand, in the same shape as `networks.json`
   -  Today that is Lens (232), BNB (56), Linea (59144) and Plasma (9745)
- `dev/generate-networks-file.sh`: 
   -  Restores the original jq pipeline over `broadcast/*/*/*.json`
   -  Obviously, adapts it to take the `broadcast/networks-manual.json` intro account
- `networks.json`:
   - is regenerated now as a result of running the script
   - only change u see there is related to the casing. The script uses loweccase

## Test plan

Regenerate and confirm the file does not have changes:

```bash
bash dev/generate-networks-file.sh > networks.json
git diff --exit-code networks.json 
```

A CI check that enforces this invariant follows in a separate PR.

## Out of scope
Probably after these 2 PRs I want to revisit the script and see if it all work together. For now, im happy the networks.json can be reconstructed again from current state of the project.
- https://github.com/cowprotocol/composable-cow/pull/142
- https://github.com/cowprotocol/composable-cow/pull/141

### DIFF
--- a/dev/generate-networks-file.sh
+++ b/dev/generate-networks-file.sh
@@ -14,15 +14,18 @@ generated=$(for deployment in "$repo_root_dir/broadcast/"*"/"*"/"*".json"; do
   # Extract contract info per chain. `CREATE` is accepted alongside `CREATE2` because the
   # standalone `deploy_ComposableCowPoller.s.sol` omits the salt, so the existing Gnosis Chain
   # poller was deployed non-deterministically.
-  jq --arg chainId "$chain_id" '
-    .transactions[]
+  jq --compact-output --arg chainId "$chain_id" '
+    # Foundry recorded seconds in older runs and milliseconds in newer ones.
+    (.timestamp | if . > 100000000000 then . / 1000 else . end | floor) as $ran_at
+    | .transactions[]
     | select(.transactionType == "CREATE" or .transactionType == "CREATE2")
     | select(.hash != null)
-    | {(.contractName): {($chainId): {address: .contractAddress, transactionHash: .hash }}}
+    | [$ran_at, {(.contractName): {($chainId): {address: .contractAddress, transactionHash: .hash }}}]
   ' <"$deployment"
-  # When the same contract appears more than once, the last one wins: within a run, the latest
-  # deployment; across runs, the path that sorts last (so `run-latest.json` beats `run-<timestamp>.json`).
-done | jq --sort-keys --null-input 'reduce inputs as $item ({}; . *= $item)')
+  # When the same contract appears more than once the most recent run wins. Ordering by run
+  # timestamp keeps that true across scripts, whose file paths sort arbitrarily. `sort_by` is
+  # stable, so a run's own transactions stay in order and the later one wins.
+done | jq --sort-keys --slurp 'sort_by(.[0]) | map(.[1]) | reduce .[] as $item ({}; . *= $item)')
 
 # Merge with manual file if it exists
 if [[ -f "$manual_file" ]]; then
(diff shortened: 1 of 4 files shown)


dev/generate-networks-file.sh
```diff
@@ -0,0 +1,39 @@
...
+  # poller was deployed non-deterministically.
+  jq --arg chainId "$chain_id" '
+    .transactions[]
+    | select(.transactionType == "CREATE" or .transactionType == "CREATE2")
+    | select(.hash != null)
+    | {(.contractName): {($chainId): {address: .contractAddress, transactionHash: .hash }}}
+  ' <"$deployment"
+  # When the same contract appears more than once, the last one wins: within a run, the latest
+  # deployment; across runs, the path that sorts last (so `run-latest.json` beats `run-<timestamp>.json`).
+done | jq --sort-keys --null-input 'reduce inputs as $item ({}; . *= $item)')
```
igorroncevic: "Last deployment wins" is actually "last script path" wins. 

#142 records a successful Poller deployment at `0xd079...` that was deployed _after_ the currently published `0x5Eda...`, but this script selects `0x5Eda...` as valid because `deploy_ComposableCowPoller.s.sol` ends up after `deploy_All.s.sol` when sorted.

Evidence:
- [deploy-latest.json](https://github.com/cowprotocol/composable-cow/blob/94a92275f4717ea6145c41723c38a1b5c5740ded/broadcast/deploy_All.s.sol/100/run-latest.json#L4-L10) from #142 
- [networks.json](https://github.com/cowprotocol/composable-cow/blob/94a92275f4717ea6145c41723c38a1b5c5740ded/networks.json#L36-L40) records the other address

I think we should sort by block/timestamp from the filename rather than path

dev/generate-networks-file.sh
```diff
@@ -0,0 +1,39 @@
...
+
+  # Extract contract info per chain. `CREATE` is accepted alongside `CREATE2` because the
+  # standalone `deploy_ComposableCowPoller.s.sol` omits the salt, so the existing Gnosis Chain
+  # poller was deployed non-deterministically.
+  jq --arg chainId "$chain_id" '
+    .transactions[]
+    | select(.transactionType == "CREATE" or .transactionType == "CREATE2")
+    | select(.hash != null)
+    | {(.contractName): {($chainId): {address: .contractAddress, transactionHash: .hash }}}
+  ' <"$deployment"
```
igorroncevic: This selects transactions by type and non-null hash without checking the corresponding receipt status. Can we require a matching receipt with [`status == "0x1"`](https://github.com/cowprotocol/composable-cow/blob/94a92275f4717ea6145c41723c38a1b5c5740ded/broadcast/deploy_All.s.sol/100/run-latest.json#L25-L31) so a logged failed broadcast cannot become the published deployment?

igorroncevic (changes requested):

dev/generate-networks-file.sh
anxolin: Fixed! good catch. Also the sorting by timestamp had to deal with 2 formats (seconds and milliseconds) for different versions in foundry

869daf0687e3f204bfaef9f9128465073aa43454

dev/generate-networks-file.sh
anxolin: 547acf5110e18c2bb1155d4a6f2d834e8334e53e

igorroncevic (approved):

Issue #93, opened 2024-10-25
bug: current contract deployments can't be replicated with current code


fedgiac:
### Problem

It doesn't appear to be possible to deterministically redeploy the contracts with the code in this repo anymore, not even by checking out old commits.

### Impact

Deterministic contract addresses are either not available or require bespoke procedures when deploying the contract on a new chain. Contract verification becomes in turn more difficult.

### To reproduce

Execute the deployment command without broadcasting:

```sh
$ ETH_RPC_URL='https://eth.llamarpc.com'
$ forge script script/deploy_ProdStack.s.sol:DeployProdStack --rpc-url $ETH_RPC_URL -vvvv
```

Extract the expected deployment addresses:
```
$ cat ./broadcast/deploy_ProdStack.s.sol/1/dry-run/run-latest.json | jq --sort-keys '[ .transactions[] | select(.transactionType=="CREATE2") | { key: .contractName, 
value: .contractAddress} ] | from_entries' 
{
  "ComposableCoW": "0x3b69dfad2957ac71f8e1ee81f6901e6cb6695f1c",
  "CurrentBlockTimestampFactory": "0x302f261eb0b22dfb4a3d29f32503b1b6f7600f5b",
  "ExtensibleFallbackHandler": "0xa610bd67f02e84d5493ed002c3bb45d0415bf0df",
  "GoodAfterTime": "0x3b9122d50c0bbb81672c355187a05b301b00ce75",
  "PerpetualStableSwap": "0x963f411ac754055b611fe464fa4d50772e9b1f9c",
  "StopLoss": "0xab83e5bcce08b7c014c287cd07b6a0d4d9399746",
  "TWAP": "0xbf39aef05ca2f9c92b01088f90ebfc0715f02b75",
  "TradeAboveThreshold": "0x6e54d4d8f52a71cc6c6d95a6f06e696f69f99abf"
}
```

Compare with current addresses and observe that they all differ:
```sh
$ cat ./networks.json | jq --sort-keys '[. | to_entries[] | {key:.key, value: .value."1".address}] | from_entries'
{
  "ComposableCoW": "0xfdaFc9d1902f4e0b84f65F49f244b32b31013b74",
  "CurrentBlockTimestampFactory": "0x52eD56Da04309Aca4c3FECC595298d80C2f16BAc",
  "ExtensibleFallbackHandler": "0x2f55e8b20D0B9FEFA187AA7d00B6Cbe563605bF5",
[...]

PR #141, opened 2026-08-06
feat: deploy the official contracts to new chains

anxolin:
A way to deploy the official contracts to a new chain at their canonical addresses, without solving #93 first.

## Why not recompile

Currently we are still affected by #93. 

Recompiling gives byte-identical *runtime* code but a different Solidity metadata hash, and that hash sits inside the initcode, so the `CREATE2` address moves. For `GoodAfterTime` on Gnosis:

Reproducing the old metadata would mean restoring the source paths of a deleted repo and freezing them forever, with no signal when it silently breaks.

## What this does instead

Records the initcode of the contracts already deployed, in `canonical/`, and replays it through the deterministic deployment proxy. 

🥳 Same initcode + same salt + same deployer gives the same address on any EVM chain, no compiler involved.

## dev/deploy-canonical.sh
It's the script used for deploying the "cannonical contracts" .

For new contracts, we use now determinstic deployments. Out of the scope to revisit foundry config, but I will make sure we use https://www.getfoundry.sh/guides/deterministic-deployments-using-create2 

## Deterministic deployer
The script depends on this deployer being present in the chain `0x4e59b44847b379578588920cA78FbF26c0B4956C` 

https://www.getfoundry.sh/guides/deterministic-deployments-using-create2#create2_deployer

The script asserts the existance of it.

## Note on salts

They are not uniform, which is why `manifest.json` records them per contract rather than assuming one. 

### 1. Dry run in gnosis
```bash
ETH_RPC_URL=<gnosis-rpc>
bash dev/deploy-canonical.sh --rpc-url $ETH_RPC_URL
```

It should give you the canonical address!
```
ComposableCoW                  0xfdaFc9d1902f4e0b84f65F49f244b32b31013b74  already deployed
CurrentBlockTimestampFactory   0x52eD56Da04309Aca4c3FECC595298d80C2f16BAc  already deployed
ExtensibleFallbackHandler      0x2f55e8b20D0B9FEFA187AA7d00B6Cbe563605bF5  already deployed
GoodAfterTime                  0xdaf33924925e03c9cc3a10d434016d6cfad0add5  already deployed
PerpetualStableSwap            0x519BA24e959E33b3B6220CA98bd353d8c2D89920  already deployed
StopLoss                       0x412c36e5011cd2517016d243a2dfb37f73a242e7  already deployed
TWAP                           0x6cF1e9cA41f7611dEf408122793c358a3d11E5a5  already deployed
TradeAboveThreshold            0x812308712a6d1367f437e1c1e4af85c854e1e9f6  already deployed
```

### 2. Dry run in a network it hasn't been deployed yet
~I used robinhookd chain. Should give you the same addresses!~
```
ComposableCoW                  0xfdaFc9d1902f4e0b84f65F49f244b32b31013b74  would deploy
CurrentBlockTimestampFactory   0x52eD56Da04309Aca4c3FECC595298d80C2f16BAc  would deploy
ExtensibleFallbackHandler      0x2f55e8b20D0B9FEFA187AA7d00B6Cbe563605bF5  would deploy
GoodAfterTime                  0xdaf33924925e03c9cc3a10d434016d6cfad0add5  would deploy
PerpetualStableSwap            0x519BA24e959E33b3B6220CA98bd353d8c2D89920  would deploy
StopLoss                       0x412c36e5011cd2517016d243a2dfb37f73a242e7  would deploy
TWAP                           0x6cF1e9cA41f7611dEf408122793c358a3d11E5a5  would deploy
TradeAboveThreshold            0x812308712a6d1367f437e1c1e4af85c854e1e9f6  would deploy
```

EDIT: 24/08/2026. Now deploying in Robinhood shows this error (which is expected)

image

### 4. Deploy in a network that is not deployed yet
I used anvil because deploying for real would require to deploy the settlement contract, dependency of composable cow.

```bash
anvil --silent --chain-id 100 &

SETTLEMENT=0x9008D19f58AAbD9eD0D60971565AA8510560ab41
cast rpc anvil_setCode $SETTLEMENT \
  "$(cast code $SETTLEMENT --rpc-url https://rpc.gnosischain.com)" \
  --rpc-url http://127.0.0.1:8545

export PRIVATE_KEY=0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
bash dev/deploy-canonical.sh --rpc-url http://127.0.0.1:8545 --broadcast
```

The result should look like:
```
Verifying recorded initcode...
All recorded initcode matches the canonical addresses.

ComposableCoW                  0xfdaFc9d1902f4e0b84f65F49f244b32b31013b74  deploying...
ComposableCoW                  0xfdaFc9d1902f4e0b84f65F49f244b32b31013b74  deployed
CurrentBlockTimestampFactory   0x52eD56Da04309Aca4c3FECC595298d80C2f16BAc  deploying...
CurrentBlockTimestampFactory   0x52eD56Da04309Aca4c3FECC595298d80C2f16BAc  deployed
ExtensibleFallbackHandler      0x2f55e8b20D0B9FEFA187AA7d00B6Cbe563605bF5  deploying...
ExtensibleFallbackHandler      0x2f55e8b20D0B9FEFA187AA7d00B6Cbe563605bF5  deployed
GoodAfterTime                  0xdaf33924925e03c9cc3a10d434016d6cfad0add5  deploying...
GoodAfterTime                  0xdaf33924925e03c9cc3a10d434016d6cfad0add5  deployed
PerpetualStableSwap            0x519BA24e959E33b3B6220CA98bd353d8c2D89920  deploying...
PerpetualStableSwap            0x519BA24e959E33b3B6220CA98bd353d8c2D89920  deployed
StopLoss                       0x412c36e5011cd2517016d243a2dfb37f73a242e7  deploying...
StopLoss                       0x412c36e5011cd2517016d243a2dfb37f73a242e7  deployed
TWAP                           0x6cF1e9cA41f7611dEf408122793c358a3d11E5a5  deploying...
TWAP                           0x6cF1e9cA41f7611dEf408122793c358a3d11E5a5  deployed
TradeAboveThreshold            0x812308712a6d1367f437e1c1e4af85c854e1e9f6  deploying...
TradeAboveThreshold            0x812308712a6d1367f437e1c1e4af85c854e1e9f6  deployed

Deployed 8 contract(s), skipped 0 already present.
Transactions recorded in broadcast/canonical/100.json; merge them into networks.json.
```

# Out of scope
This PR only create the deterministic canonnical deployments to address #93. A follow up PR will improve the deployment scripts so it becomes simpler and more deterministic https://github.com/cowprotocol/composable-cow/pull/142


dev/deploy-canonical.sh
```diff
@@ -0,0 +1,135 @@
...
+  if [[ "$(code_at "$address")" != "0x" ]]; then
+    printf '%-30s %s  already deployed\n' "$name" "$address"
+    skipped=$((skipped + 1))
+    continue
+  fi
+
+  if [[ "$broadcast" != true ]]; then
+    printf '%-30s %s  would deploy\n' "$name" "$address"
+    continue
+  fi
```
igorroncevic, line 152: The dry run can print that `ComposableCoW` “would deploy” without checking that `GPv2Settlement` is already deployed. `ComposableCoW`’s constructor immediately calls `GPv2Settlement.domainSeparator()`, so the real deployment will revert the code is missing at that address. 

Can we detect this during preflight and print a clear error?

dev/deploy-canonical.sh
```diff
@@ -0,0 +1,135 @@
...
+
+  printf '%-30s %s  deploying...\n' "$name" "$address"
+  cast send "$deployer" "0x${salt#0x}${initcode#0x}" \
+    --rpc-url "$rpc_url" --private-key "$PRIVATE_KEY" >/dev/null
```
igorroncevic: Redirecting cast send to `/dev/null` loses the deployment transaction hash. `networks.json` needs both the address and transaction hash, so human must recover it manually later. 

Can we write output to some file like `networks.temp.json`?

canonical/manifest.json
```diff
@@ -0,0 +1,42 @@
+{
+  "ComposableCoW": {
+    "address": "0xfdaFc9d1902f4e0b84f65F49f244b32b31013b74",
+    "salt": "0x76312e302e300000000000000000000000000000000000000000000000000000",
+    "saltString": "v1.0.0"
```
igorroncevic, line 5: I don't see the `saltString` being used anywhere, do we remove it or include it?

EDIT: is it consumed by `DeployScript` in #142 ?

dev/deploy-canonical.sh
```diff
@@ -0,0 +1,135 @@
...
+  expected=$(jq -r --arg n "$name" '.[$n].address' "$manifest")
+  initcode=$(cat "$initcode_dir/$name.initcode")
+  computed=$(cast create2 --deployer "$deployer" --salt "$salt" --init-code "$initcode" | tail -1)
+  if [[ "$(tr 'A-Z' 'a-z' <<<"$computed")" != "$(tr 'A-Z' 'a-z' <<<"$expected")" ]]; then
+    echo "Error: $name initcode does not produce its canonical address." >&2
+    echo "       expected $expected, computed $computed" >&2
+    exit 1
+  fi
+done < <(jq -r 'keys[]' "$manifest")
+echo "All recorded initcode matches the canonical addresses."
```
igorroncevic: Could we run the manifest verification in CI? A small step can recompute every CREATE2 address from its salt and initcode, then compare it with manifest.json. This needs no RPC or private key and would catch accidental artifact changes when the PR is opened.

e.g.
```
  - name: Verify canonical deployment addresses
    shell: bash
    run: |
      deployer=0x4e59b44847b379578588920cA78FbF26c0B4956C
      manifest=canonical/manifest.json

      while IFS= read -r name; do
        salt=$(jq -r --arg name "$name" '.[$name].salt' "$manifest")
        expected=$(jq -r --arg name "$name" '.[$name].address' "$manifest")
        initcode=$(<"canonical/initcode/$name.initcode")

        computed=$(cast create2 \
          --deployer "$deployer" \
          --salt "$salt" \
          --init-code "$initcode")

        if [[ "${computed,,}" != "${expected,,}" ]]; then
          echo "::error file=canonical/manifest.json::$name: expected $expected, computed $computed"
          exit 1
        fi
      done < <(jq -r 'keys[]' "$manifest")
```

igorroncevic (changes requested):

canonical/manifest.json
anxolin: Yes, its not. Only address/salt

its just documentaton, I could read it in the script and compare to the salrt, but probably overkill. I expect we will kill this `canonical` dir the first chance we get.

dev/deploy-canonical.sh
anxolin: Makes sense to add something for sloppy hands. Initially thought might be a bit overkill, cause we will kill `canonical` as soon as we can and I don't think we will add any contract there  (so requires no mantainance). 

If this breaks, is visible by the script deploying a different contract than expected when we go to a new chain, so its noticable and easily recoverable (can only happen if someone changes the init codes or sth, so is visible in git). 

Anyways, I think is easy to add (with claudes help :P) 

I moved the check to its own script: cd35e6029f59951757fc7832bf640d3d726bc567

Also I added a GH action, which as u point out, don't need RPC, so it was a fairly easy check: 3c6b2c55ae63844257ec885e3370fbb5adc4e8f5

dev/deploy-canonical.sh
anxolin: Added c6d1e3a, although I think other than giving nicer feedback matters too much. It would still fail

when deploying it tries to estimate gas, and it would already fail there

dev/deploy-canonical.sh
anxolin: O boi, I just wanted to do some simple script to deploy the contracts for the legacy deployments, I was not intending to do a fully featured script. But good call, if added to the script the feature to write the deployment so the networks.json can be generated automatically from it

421349a

dev/deploy-canonical.sh
```diff
@@ -0,0 +1,162 @@
...
+if [[ -n "$records" ]]; then
+  mkdir -p "$deployment_dir"
+  existing='{}'
+  if [[ -f "$deployment_file" ]]; then
+    existing=$(cat "$deployment_file")
+  fi
+  printf '%s' "$records" | jq --slurp --sort-keys --arg chain "$chain_id" --argjson existing "$existing" '
+    map({(.name): {($chain): {address: .address, transactionHash: .transactionHash}}})
+    | reduce .[] as $record ($existing; . * $record)
+  ' > "$deployment_file"
```
igorroncevic, line 183: This saves deployment results as `broadcast/canonical/<chain>/deployment.json`. That location matches #139’s `broadcast/*/*/*.json` search, but #139 expects every matching file to be a Foundry artifact containing timestamp, receipts, and transactions. 

This new file has a different structure, so regeneration fails with `Cannot iterate over null`. Can we move it outside that search path or update #139 to handle this format?

anxolin: Addressed in 2 parts; one is aiming to make this PR green https://github.com/cowprotocol/composable-cow/pull/141/changes/6d25adff0235269ebc318689b29adf1a9e39ddc4

The other is fix in the script that generates the networks.json to not pick the canonnical jsons (only foundry `run-*.json` files) https://github.com/cowprotocol/composable-cow/pull/144

dev/deploy-canonical.sh
```diff
@@ -0,0 +1,164 @@
...
+settlement="0x${composable_cow_initcode: -40}"
+if [[ "$(code_at "$settlement")" == "0x" ]]; then
+  cat >&2 <<EOF
+Error: GPv2Settlement is not present at $settlement on chain $chain_id.
+
+ComposableCoW's constructor calls domainSeparator(), so it must be deployed first.
+EOF
+  exit 1
+fi
+echo "GPv2Settlement: present"
```
igorroncevic: Could we verify the Settlement’s `domainSeparator()` against the value expected for this chain and address, instead of only checking that some code exists?

canonical/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+This folder keeps the initcode of the deployments that already exist, read back from their creation
+transactions: one file per contract under `initcode/`, plus `manifest.json` recording each salt and
+resulting address.
+
+Deploy it with [`dev/deploy-canonical.sh`](../dev/deploy-canonical.sh), which checks every file
+against `manifest.json` before sending anything. The target chain needs the
+[deterministic deployment proxy](https://github.com/Arachnid/deterministic-deployment-proxy) at
+`0x4e59b44847b379578588920cA78FbF26c0B4956C`, since its address is part of the `CREATE2` preimage;
+the script stops if it is missing.
```
igorroncevic: `GPv2Settlement` is also a mandatory prerequisite, but this setup section only mentions the deterministic deployment proxy. Can we document the required Settlement address here too?

dev/deploy-canonical.sh
anxolin: Addressed 73ce23b

canonical/README.md
anxolin: Addressed 73ce23b

dev/deploy-canonical.sh
igorroncevic: IIUC now records are written only after every deployment finishes. Any later failure loses earlier transaction hashes. Should we save each receipt immediately?

igorroncevic (approved):
LGTM 👍🏻

PR #142, opened 2026-08-06
refactor: rework the deployment scripts and pin them to deterministic settings

anxolin:
Follow-up to #141 

Makes deployment simpler and more predictable. This is done by adding the following changes

## 1. Simplify deployment of all contracts
After #141 a normal deployment of the contracts is done by
1. Deploy legacy contracts (`ComposableCoW`, etc.)
2. Deploy new contracts (i.e the poller)

The scripts were a bit confusing, because
- `script/deploy_ProdStack.s.sol` would deploy to addresses that are not what you expect! Also `ProdStack` gives me little information, not sure if only the legacy contracts should be there, or is aiming to deploy all contracts.
- `script/deploy_OrderTypes.s.sol` is weird, deploys a lot of the contracts deployed in `ProdStack` script already. 

First thing I did is to delete both scripts, and replace it with a `script/deploy_All.s.sol` which deploys everything (legacy and new).

## 2. Default to not deploy legacy
The README suggest you to deploy the legacy contracts using the script `dev/deploy-canonical.sh`

So if `script/deploy_All.s.sol` deploys these contracts deployment is confusing. This is what was happening. 

Now the script won't deploy by default the legacy contracts (like `ComposableCow`). So deploying all contracts now is a simple run of the 2 scripts!

## 3. Allow to re-deploy the legacy contracts too
It is convenient to also be able to deploy the contracts again (not in the canonical address, but in the one that would come natuarlly given the CREATE2 setup of this repo).

`script/deploy_All.s.sol` allows to pass an environment variable `CANONICAL=false` which will redeploy all legacy contracts into the address that would match this project's code. 

## 4. Salt refactor
All contracts deployed using the same salt over and over (`v1.0.0`). Code was not DRY!

This has been:
- Refactored, so its defined in a single place https://github.com/cowprotocol/composable-cow/pull/142/changes#diff-c4d39d13ff2a63faacda1b9fc857d69773fa702609bf165fa31dd740cdb9de52R10
- Allows to override by env: Handy to deploy a new set of contract with a new SALT

## 5. Incoherence with deployment of individual contract
Before this PR `script/deploy_ProdStack.s.sol` would deploy all contracts using `CREATE2`, but the individual deployments used `CREATE`.

This PR make it coherent, and reuses the salt, so using `script/deploy_All.s.sol` or `script/deploy_ComposableCoW.s.sol` will give you the same address and they will both use the salt defined in your env. 

## 6. Deterministic deployments
[...]

PR #143, opened 2026-08-06
ci: check networks.json drift

anxolin:
Follow-up to #139.

Checks for drift in networks.json

The idea is to prevent to merge PRs with drift.

## Test plan

Simulating the job against #139's tree passes:

```bash
bash dev/generate-networks-file.sh > networks.json
git diff --exit-code networks.json
```

To watch it fail, change any address in `networks.json` and run the same two commands


igorroncevic (approved):
This could cause conflicts with #141 and #142, perhaps we should rebase this on top of them.

PR #144, opened 2026-08-13
fix: read only Foundry run artifacts when regenerating networks.json

anxolin:
Follow-up to #139 motivated by this comment https://github.com/cowprotocol/composable-cow/pull/141#discussion_r3772049100.

## How to test

Verify a regeneration of the networks would change nothing in `networks.json`
```
bash dev/generate-networks-file.sh | diff - networks.json
```

Create a empty json to mess up with the `generate-networks-file.sh`
This replicates what #141 is doing (adding a json with a different format there)

```
mkdir -p broadcast/canonical/999 && echo '{}' > broadcast/canonical/999/deployment.json
```

You can test this script breaks in `main`:
```bash
git show origin/main:dev/generate-networks-file.sh | bash 
```

Same test, should work in this branch, and should not create any changes in the `networls.json` file:
```
bash dev/generate-networks-file.sh | diff - networks.json 
```

Cleanup after your test:
```
rm -rf broadcast/canonical
```

### DIFF
--- a/dev/generate-networks-file.sh
+++ b/dev/generate-networks-file.sh
@@ -7,6 +7,8 @@ manual_file="$repo_root_dir/broadcast/networks-manual.json"
 
 # One record per deployed contract, oldest first, so the newest run wins the merge below. Ordering
 # by run rather than by file path matters because several scripts deploy the same contract.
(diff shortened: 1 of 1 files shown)


igorroncevic (approved):
Should we rebase it on top of #142 ?

PR #145, opened 2026-08-14
feat(poller): register a schedule from the funder's CowShed

anxolin:
## What and why
I will explain the motivation of this PR through a series of assumptions and conclusions

**Assumption 1**: 
- To trade using cow-shed, the founder already needs to sign the ERC-712 message  for the approval of the vault relayer and creation of the TWAP. 
- So, the founder needs to sign a CowShed bundle (call to `shed.executeHooks(calls)`). 

**Assumption 2**: 
- The founder, in its first TWAP, needs to do the approval to the poller, so the funds can be polled JIT. 
- Normally this is a tx, or could be a permit executed as part of the sell=buy order.

**Assumption 3**: 
- The founder needs to register the schedule in the poller contract. 
- This can be a TX or a 712 message, but it's a different signature than the one required for "assumption 1"

----- 
**Conclusion 1**: 
- If we allowed the poller contract to register on behalf of its owner, we could bundle the registration as part of "Assumption 1".
- This is easy to do, because given a shed is easy to get the owner, so its easy to validate the shed only register on behalf of owner's account

**Conclusion 2**: 
- Due to "Conclusion 1" After we approve the poller, we can bundle all operations needed to create a TWAP in a single call to `shed.executeHooks(calls)`. 
- So our sell=buy order could just have a single hook instead of 2, and we need one signature less.

**Conclusion 3**: 
- As "Conclusion 2" concluded, sell=buy needs just one hook. This means we could replace the order creation with a single transaction.  This simplifies the flow drastically. 
- Wallets can simulate all the creation, you don't rely in some obscure side effect of a non-operation order

**Conclusion 4**: 
- If the token is permitable, we could include the permit call as part of `shed.executeHooks(calls)`. 
- This would allow to further reduce to one TX the (1) approval of the poller, (2) the schedule registration, (3) the vault relayer approval (4) the TWAP creation

## Changes
Adds a new function `registerFromShed` which allows cow-shed to register on behalf of its owner. 

Adds a dependency with cow-shed in the poller, which in principle I don't love, but:
- There's a strong motivation do do so. Enables a really simpler flow!
- It's opt-in, the poller can be used by non cow-shed flows, so you can still register by calling to the `register` function, or its counterparty that takes a 712 signature.

## PoC
Verify this transaction, which generated the full TWAP setup with a single TX. 
https://app.blocksec.com/phalcon/explorer/tx/gnosis/0x7964ac660ab6a76d3e7c45db41ccbc21ebc1bc91ae3fd678c860f97b09148903

How it works: We only sign a TX to cow-shed to execute 4 calls

- EOA: set allowance for the poller via permit. One time only, next TWAPs can skip this call. Requires a signature.
- Cow-shed: Register schedule
- Cow-shed: Set allowance for vault relayer
- Cow-shed: Create TWAP

image

## Pros/cons to use a single TX instead of sell=buy
After this PR it changes my recommendation to use sell=buy for TWAP for EOAs. Now I'm in favour of using a single TX like the one in the PoC.

Good: 
- Obviously, way less signatures!! (main motivation)
- Simpler, you just do a direct execution of the TWAP creation & setup, instead of using the indirection of a sell=buy (hack)
- It's simpler to review the TX, because technically wallets could simulate the TX u execute (sell=buy is more obscure, which makes it more risky)
- Easier to explain the UI to the user

Bad: 
- Is not gasless, but anyways, you can opt out and use other mechanisms

## Test

```bash
```


igorroncevic (commented):
Must rebase first to include latest changes in order to be reviewable.

src/types/ComposableCowPoller.sol
```diff
@@ -229,6 +273,30 @@ contract ComposableCowPoller is EIP712 {
...
+    ///      schedule live.
+    function revokeFromShed(IConditionalOrderGenerator handler, address funder, address owner, bytes32 salt)
+        external
+        returns (bytes32 id)
+    {
+        _requireFunderShed(funder);
+
+        id = _scheduleId(funder, handler, owner, salt);
+        _revoke(id, funder, owner, schedules[id].authEpoch);
+    }
```
igorroncevic, line 359: `revokeFromShed` doesnt check the epoch authorized by the CowShed bundle, so delayed old revocation can delete a replacement schedule in a newer epoch. Please pass and validate the expected `authEpoch`, like `revokeWithSignature`, and test it

script/deploy_AnvilStack.s.sol
```diff
@@ -47,6 +47,8 @@ contract DeployAnvilStack is Script {
 
     function run() external {
         uint256 deployerPrivateKey = vm.envUint("PRIVATE_KEY");
+        address cowShedFactory = vm.envAddress("COW_SHED_FACTORY");
```
igorroncevic: Please update `.env.example` with this value

test/ComposableCowPoller.t.sol
```diff
@@ -807,4 +819,341 @@ contract ComposableCowPollerTest is BaseComposableCoWTest {
...
+        vm.expectRevert(ComposableCowPoller.InvalidAuthEpoch.selector);
+        poller.registerFromShed(schedule);
+    }
+
+    // --- end to end -------------------------------------------------------------------------
+
+    /// @dev The flow this change exists for: one funder signature over a shed bundle that both
+    ///      creates the TWAP and registers the schedule, then just-in-time funding per part with no
+    ///      further signature. No ERC-1271 is needed here because the shed creates the order itself.
+    function test_registerFromShed_endToEndSignedBundle() public {
```
igorroncevic, line 1174: This flow still needs ERC-1271 because the Shed owns the order, but the mock doesnt implement it. Please test signature validation using the real `COWShedForComposableCoW`, otherwise registration and funding can succeed while settlement fails.

igorroncevic (changes requested):

src/types/ComposableCowPoller.sol
anxolin: Yes, makes sense. 

For a moment I was thinking if this should just kill the last version regardless of its version (kind of making it indempotent revocation). 

I felt this case is not the same as the signed ERC-712 flows, because the signature can easily be leaked, so someone could just use it to revoke over and over new schedules. 

Here is the account making a transaction to make sure there's no schedule. 

Anyways, I will do as you say, because makes sense to be excplicit with the version u want to revoke, and if this is used in for example a loose hook in an order you forgot about, u don't run into race conditions.

anxolin: 95a6186bcd55781cff4fbb8fd3dcfb1f01a58f6a

test/ComposableCowPoller.t.sol
anxolin: 468f4acf590c62304c4edeccc864f32666b491ce

.env.example
```diff
@@ -14,6 +14,9 @@ SETTLEMENT=0x9008D19f58AAbD9eD0D60971565AA8510560ab41
 # ComposableCoW (canonical)
 COMPOSABLE_COW=0xfdaFc9d1902f4e0b84f65F49f244b32b31013b74
 
+# CowShed factory (canonical)
+COW_SHED_FACTORY=0x00E989b87700514118Fa55326CD1cCE82faebEF6
```
igorroncevic, line 18: IIUC this is the old factory that doesn't deploy ERC-1271 compatible sheds? Shouldn't we use `COWShedForComposableCoW` (`0x4F4350bf2c74aaCD508D598a1ba94EF84378793d`)?

src/types/ComposableCowPoller.sol
```diff
@@ -25,6 +33,13 @@ contract ComposableCowPoller is EIP712 {
     /// @dev `ComposableCoW` stores the settlement domain separator supplied at deployment.
     ComposableCoW public immutable COMPOSABLE_COW;
 
+    /// @notice The one CowShed factory whose proxies may manage schedules for their own owner.
+    /// @dev Pinned at deployment. It must be a factory whose only deployment path derives the proxy
+    ///      from the owner alone, so that a call from `proxyOf(funder)` proves `funder` authorised
+    ///      it. See `registerForFunder`. It must also deploy `COWShedForComposableCoW` sheds: a shed
```
igorroncevic: nit: update to `registerFromShed`

src/types/ComposableCowPoller.sol
```diff
@@ -64,6 +79,13 @@ contract ComposableCowPoller is EIP712 {
     /// @notice Thrown when someone other than the schedule funder registers, updates, or revokes a schedule.
     error OnlyFunder();
 
+    /// @notice Thrown when a `ForFunder` action does not come from the funder's own CowShed, or when
```
igorroncevic: nit: there's a few places where `ForFunder` is mention, and it should be `ForShed`

README.md
```diff
@@ -122,6 +122,8 @@ Fortunately, when using Safe, it is possible to batch together all the above cal
 
 The funder can register or revoke a funding schedule directly. To avoid a separate transaction, the funder can instead sign an EIP-712 authorization that a relayer submits on-chain. The Poller validates EOA signatures directly and uses ERC-1271 when the funder is a contract. Signed actions include a deadline and the current authorization epoch. Revocation derives the schedule ID from its funder, handler, owner, and salt, so it works before or after registration without allowing one funder to cancel another's schedule. It advances the authorization epoch, invalidating pending signatures from prior epochs while allowing the same schedule ID to be reused.
 
+A funder that drives its orders through its [CowShed](https://github.com/cowprotocol/cow-shed) can skip the signature entirely: `registerFromShed` and `revokeFromShed` accept calls from `proxyOf(funder)` on the factory pinned at deployment, which already proves the funder authorized them. The schedule ID stays namespaced by the funder, so the funder keeps unilateral `revoke` over whatever its shed registered.
```
igorroncevic: nit: don't we technically only skip the separate poller signature?

README.md
igorroncevic: nit: add `COW_SHED_FACTORY`

README.md
igorroncevic: nit: it should be `https://github.com/cowdao-grants/cow-shed`

igorroncevic (dismissed):
soft approve - mostly outdated stuff and typos, feel free to merge after resolving

.env.example
anxolin: Good point, but I think we should use `2.1.0` which is the latest (0x4F4350bf2c74aaCD508D598a1ba94EF84378793d  points  to cow-shed `2.0.0`)

anxolin: This v2.1.0 comes from a draft PR, I will deal with it after we freeze this repository https://github.com/cowdao-grants/cow-shed/pull/79

src/types/ComposableCowPoller.sol
anxolin: dbd6bca1741b937d31ddde4776430d16b0d2aec0

src/types/ComposableCowPoller.sol
anxolin: b7e1cf7917469242022fb5a2446a84ab8c7bdba8

README.md
anxolin: Tried to write in a different way fc5fc4d49665898f997e7077b18d2d18d2367502

README.md
anxolin: fc5fc4d

igorroncevic (approved):

PR #146, opened 2026-08-24
feat(poller): add signed schedule revocation

igorroncevic:
# Description

Funders can revoke polling schedules through relayers with an EIP-712 signature. This forwards the reviewed PR #136 change, which was accidentally merged into its stacked feature branch, onto `main`.

# Changes

- [x] Add signed schedule revocation for EOA and ERC-1271 signatures.
- [x] Bind revocation authorization to the schedule ID, funder, deadline, and auth epoch.
- [x] Preserve the accepted PR #135 follow-up fixes required by the final implementation.

## How to test

1. Run `forge test --offline --match-contract ComposableCowPollerTest`.
2. Confirm all 38 tests pass.

Related: #136

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,14 @@ Fortunately, when using Safe, it is possible to batch together all the above cal
 **TODO**
 **NOTE:** For canceling a TWAP order, follow the instructions at [Conditional order cancellation](#Conditional-order-cancellation).
 
+## Just-in-time funding with `ComposableCowPoller`
+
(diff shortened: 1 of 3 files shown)


fedgiac (approved):
This is an "inherited approval" from Anxo's approval of #136.
I didn't read the code but I checked that it's line-by-line identical to #136. My understanding is that that PR should have been merged on `main` and not on the other branch. I also very briefly checked the commits between Anxo's approval and the merge,

PR #148, opened 2026-08-25
fix: bound test_simulate_fuzz by the span it actually simulates

anxolin:
The `test_simulate_fuzz` is a bit overkill of a test. 

This PR is not about rewriting it, but just fix some issue that it was making it to run out of memory (`MemoryOOG` error)

Changing the bound condition helps. Specifically when `span <> 0`, since it was bounded to `numParts * span + numParts * (frequency - span) * 3`, which does not constrain the loop. 
Large frequencies therefore ran thousands of iterations and ran out of memory.

Bounding `numParts * frequency` in both cases fixes it.

CI has skipped the test since 2023 (`--no-match-test "fork|simulate"`), which is why this was never caught.

This PR brings it back!

## Test plan

```sh
forge test --match-test test_simulate_fuzz --fuzz-seed 672679878
forge test
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,7 +33,7 @@ jobs:
 
       - name: Run Forge tests
         run: |
-          forge test -vvv --optimize --optimizer-runs 20000 --no-match-test "fork" --fuzz-seed 672679878
+          forge test
         id: test
 
   networks-json:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,7 +33,7 @@ jobs:
 
       - name: Run Forge tests
         run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -33,7 +33,7 @@ jobs:
 
       - name: Run Forge tests
         run: |
-          forge test -vvv --optimize --optimizer-runs 20000 --no-match-test "fork|simulate" --fuzz-seed 672679878
```
anxolin, line 36: This can be greately simplified because
- fork: we don't have those
- simulate: main thing of this PR. the fuzz test doesnt need to be ignored now
- ` --fuzz-seed 672679878` required to make it deterministic, but moved to the foundry config so we use it always like that and we can simplify the CI

igorroncevic (approved):

PR #155, opened 2026-08-25
feat(poller): Deploy poller to Mainnet, Gnosis and Sepolia

anxolin:
This PR deploys the poller contract. This is deploying the pre-audit state to unlock the frontend. 

## Explicit factory flavoured dependency
Technically the poller depends on any cow-shed factory, however in practice, for now we need it to depend on the `COWShedFactoryForComposableCoW`, because for now we just use it for this flow. 

This is why this PR changes the env so is more explicit:

`COW_SHED_FACTORY` becomes `COW_SHED_FACTORY_FOR_COMPOSABLE_COW`.

The default points to the 2.1.0 `0x5e284e80f3bd6a7d80a8500d9c49878028110848` in https://github.com/cowdao-grants/cow-shed/blob/main/networks.json

## Drops some old deployemnts
During PoC phase we did some deployments that are not relevant any more, so the `ComposableCowPoller` record in `networks.json` is removed (and the corresponding broadcast tx)

## Uses cancun compiler
Deploy runs now need `--evm-version cancun`. The factory uses transient storage, this is what motivated the dependent PR https://github.com/cowprotocol/composable-cow/pull/156

Legacy contracts still point to the cannonical address. 

## Deployment to some networks
I deployed to 3 networks so we can propertly test: Gnosis, Mainnet and Sepolia. 

I didn't do the rest because the contracts might change during the audit, so these 3 is a good balance.

## Test plan

Make sure tests pass
```sh
FOUNDRY_PROFILE=ci forge test
```

Verify the canonical networks are untouched:
```sh
bash dev/verify-canonical.sh
```

Verify the networks files generate the commited `network.json`
```bash
diff -u networks.json <(bash dev/generate-networks-file.sh)
```


networks.json
```diff
@@ -33,12 +33,6 @@
       "transactionHash": "0xa4db8e5f949f39af60460fc05979b363b01570970e94eb8397dc39cfbdcaed86"
     }
   },
-  "ComposableCowPoller": {
-    "100": {
-      "address": "0x5eda08425781c2c39a28faaf963c79487dc91bb1",
-      "transactionHash": "0x432c451276714eacd524e2578afa05917bb6efbfdba386f1fa644cd58ab56cc3"
-    }
-  },
```
anxolin, line 41: removed since this is old deployment used in the PoC. The new is deployed in this PR

igorroncevic (approved):

PR #158, opened 2026-08-31
docs(poller): document the funder-controls-owner trust model

anxolin:
Documents the poller's trust model where the owner is picked, and pins the behaviour with a test.

`funded` is keyed by order digest, so a handler that resolves its start time from the cabinet — a TWAP with `t0 == 0` — restarts funding from the schedule's first discrete order whenever the owner re-runs `createWithContext`. The only bound is the ERC-20 allowance, which is per `(funder, sellToken)` and shared across that funder's schedules rather than a per-schedule cap.

Both are expected: `pollFunds` funds each discrete order without checking that the previous one settled, so the owner is trusted with whatever the allowance permits either way. That precondition just was not written down — the natspec on `Schedule.owner` read as permission to pick any address, where in practice it should be a CowShed the funder controls, as `registerFromShed` already enforces.

Comments only: with `bytecode_hash = "none"` and `cbor_metadata = false`, the bytecode and every CREATE2 address are unchanged.

## Test plan

- New test — funds the first part, settles it, warps 1s, re-runs `createWithContext`, asserts the same part is pulled again under a different digest. The warp is +1s rather than a whole `FREQ`, since a `FREQ` shift yields the digest the second part would have had anyway and would look like ordinary progression.

      forge test --match-test test_pollFunds_refundsFirstPartAfterCabinetRestart -vv

- Full suite, 138 passing:

      forge test

- Bytecode identical to `main` (`42deed289e018dbff72b4d6e2850b36f0bdcbf54a565fb17b86944fefcfca71c` on both):

      forge build && jq -r '.bytecode.object' out/ComposableCowPoller.sol/ComposableCowPoller.json | shasum -a 256

### DIFF
--- a/src/types/ComposableCowPoller.sol
+++ b/src/types/ComposableCowPoller.sol
@@ -130,10 +130,10 @@ contract ComposableCowPoller is EIP712 {
         bytes32 paramsHash
     );
 
-    /// @notice Emitted when a part's funds are moved.
-    /// @dev `orderDigest` is indexed so a single leg can be looked up directly, rather than only
-    ///      by walking a schedule's history. It is the CoW order struct hash, so it joins to
-    ///      settlement data.
+    /// @notice Emitted when a discrete order's funds are moved.
+    /// @dev `orderDigest` is indexed so a single discrete order can be looked up directly, rather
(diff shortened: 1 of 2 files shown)


igorroncevic (approved):
Looks good!

PR #159, opened 2026-09-09
docs: update NatSpec and README.md

igorroncevic:
# Description
Updates the NatSpec and README according to the audit report.


src/types/ComposableCowPoller.sol
```diff
@@ -60,8 +60,8 @@ contract ComposableCowPoller is EIP712 {
         address funder;
         /// @notice The address that owns the ComposableCoW conditional order and receives the pulled funds.
         /// @dev It can be an EOA or contract and may be the same address as `funder`, but the funder
-        ///      must control it: see the trust model on `pollFunds`. `registerFromShed` enforces
-        ///      that; `register` accepts any owner and leaves it to the funder.
+        ///      must trust it: see the trust model on `pollFunds`. `registerFromShed` requires the
+        ///      owner to be the funder's CowShed; `register` accepts any owner and leaves it to the funder.
```
anxolin: But this is cow-shed specific. poller could be using cow-shed or could be using something else, no?

README.md
```diff
@@ -118,13 +118,31 @@ Fortunately, when using Safe, it is possible to batch together all the above cal
...
 
-Removing the conditional order from ComposableCoW stops further funding. Revoking its funding schedule does too, but does not remove the funder's token allowance; revoke that allowance separately when it is no longer needed.
+Removing the conditional order from ComposableCoW only pauses funding. The owner can reauthorize the same parameters and resume funding while the schedule and allowance remain active. Revoking the schedule stops it, but leaves the token allowance active; revoke that allowance separately when it is no longer needed.
+
+### Handler compatibility
+
+The five handlers below use `BaseConditionalOrder`. The pre-hook first calls `getTradeableOrder` with empty `offchainInput` and funds the owner. CoW then calls `verify` with the input stored in the submitted signature; `BaseConditionalOrder` regenerates the order and compares its hash. Compatibility requires the first call to succeed and funding not to change the regenerated order.
```
anxolin: should this start mentioning that anyone can build a handler themselves and explain that poller compatibility should be taken for granted. Not all handlers will be designed to work with the poller. 

Then is when you give some examples on some handers that are defined in Composable cow

README.md
```diff
@@ -118,13 +118,31 @@ Fortunately, when using Safe, it is possible to batch together all the above cal
...
+
+The Poller has no handler allowlist. The funder must trust the handler, which can pull any token and amount covered by the allowance, and the owner, which receives the funds and controls the order authorization and cabinet context.
```
anxolin: should we mention that this is to allow anyone to make their own handler and reuse the poller (if compatible)

anxolin: Maybe we should mention some sources of incompatibility

One kind of incompatibility is when a handler can have many similar orders open at the same time, but only one can be executed. 

For example, an order whose `sellAmount` could be picked within a range of values, the contract would accept to trade for all the values in that range (yields N valid orders), this allows the poller to poll N times, potentially consuming the whole allowance. 

There might be more reasons

README.md
```diff
@@ -118,13 +118,31 @@ Fortunately, when using Safe, it is possible to batch together all the above cal
...
+
+| Handler | Poller compatibility | Reason |
+| ------- | -------------------- | ------ |
+| `TWAP` | Compatible | Each part has a fixed token and amount, accepts empty `offchainInput`, and enforces its time window. |
```
anxolin: The reason is because the contract only allows one order per part. So you can't poll twice for the same leg. 

One caveat, for a leg that doesn't execute, someone could poll the funds. Not sure if we documented such limited griefing attacks.

README.md
```diff
@@ -118,13 +118,31 @@ Fortunately, when using Safe, it is possible to batch together all the above cal
...
+| `StopLoss` | Compatible | Its token, amount, and expiry are configured; empty `offchainInput` works and oracle values only allow or reject trading. |
```
anxolin: But the reason is because yields always the same order always, no? this description is confusing to me https://github.com/cowprotocol/composable-cow/blob/main/src/types/StopLoss.sol#L106

README.md
```diff
@@ -118,13 +118,31 @@ Fortunately, when using Safe, it is possible to batch together all the above cal
...
+| `TradeAboveThreshold` | Not compatible | `sellAmount` is the owner's balance, which funding changes before verification. |
```
anxolin: I just realised we assumed the JIT contract is not compatible with some of the handler, like take TradeAboveThreshold as an example

Technically is true that today's watch-tower won't work with this. Because the order decides on the sellAmount based on the balance at the time. Balance is 0 (polling is only done at trading time).

However, a modified watch-tower that applies the hooks and interrogates the contract, will get the right order data, with the JIT funds polled.

Similarly the API will accept the order (they also simulate the hooks, so balance is there, so the handler will see the order sellAmount matches the balance)

Lastly the order would go through. Hooks are executed in the preinteractions, which is executed before the order is authorised, so funds will be there by the time 1271 validation is done.

I don't plan to enhance the watch-tower, but just saying it could be fairly easy be adapted so it works.

src/types/ComposableCowPoller.sol
igorroncevic: Of course, using cow-shed is optional, but the comment mentions `registerFromShed` specifically though.

README.md
igorroncevic: resolved in [fb48fbf](https://github.com/cowprotocol/composable-cow/commit/fb48fbf219d89a311dbe50bbddcda999747d7d76)

README.md
igorroncevic: resolved in [fb48fbf](https://github.com/cowprotocol/composable-cow/commit/fb48fbf219d89a311dbe50bbddcda999747d7d76)

README.md
igorroncevic: resolved in [fb48fbf](https://github.com/cowprotocol/composable-cow/commit/fb48fbf219d89a311dbe50bbddcda999747d7d76)

README.md
igorroncevic: resolved in [fb48fbf](https://github.com/cowprotocol/composable-cow/commit/fb48fbf219d89a311dbe50bbddcda999747d7d76)

README.md
igorroncevic: resolved in [3eee010](https://github.com/cowprotocol/composable-cow/commit/3eee0106c2babd589d875e1aeac01ddf34cfa396)

src/types/ComposableCowPoller.sol
anxolin: Fair enough, but the way its written is not clear `registerFromShed` is the only option. Maybe make it more explicit.

"If the owner is a cow-shed, you may register using `registerFromShed` which already checks the admin of cow-shed is the funder address.

README.md
```diff
@@ -118,13 +118,35 @@ Fortunately, when using Safe, it is possible to batch together all the above cal
...
+
+| Handler               | Poller compatibility                    | Reason                                                                                                                      |
+| --------------------- | --------------------------------------- | --------------------------------------------------------------------------------------------------------------------------- |
+| `TWAP`                | Compatible                              | Each part has a fixed order for an unchanged start time, accepts empty `offchainInput`, and enforces its time window.      |
+| `StopLoss`            | Compatible                              | Returns the same order whenever its checks pass and accepts empty `offchainInput`.                                        |
+| `TradeAboveThreshold` | Not supported by the current watchtower | `sellAmount` is the owner's balance, which funding changes before verification.                                             |
```
anxolin: I thought we concluded it is compatible. You could hint that if the watch-tower can guess the balance (for example including the hooks in the simulation, before polling the handler for the order), then it would work

README.md
```diff
@@ -118,13 +118,35 @@ Fortunately, when using Safe, it is possible to batch together all the above cal
...
+| `PerpetualStableSwap` | Not supported by the current watchtower | The sell token and amounts depend on the owner's balances, which funding changes before verification.                       |
```
anxolin: You can just say is a similar case as above for the same reason

README.md
```diff
@@ -118,13 +118,35 @@ Fortunately, when using Safe, it is possible to batch together all the above cal
...
+
+The funder must trust the handler, which can pull any token and amount covered by the allowance, and the owner, which receives the funds and controls the order authorization and cabinet context. A handler that returns different order digests for the same intended trade can trigger repeated funding, potentially exhausting the allowance.
+
+| Handler               | Poller compatibility                    | Reason                                                                                                                      |
+| --------------------- | --------------------------------------- | --------------------------------------------------------------------------------------------------------------------------- |
+| `TWAP`                | Compatible                              | Each part has a fixed order for an unchanged start time, accepts empty `offchainInput`, and enforces its time window.      |
```
anxolin: why mentioning the empty `offchainInput`?

README.md
```diff
@@ -118,13 +118,35 @@ Fortunately, when using Safe, it is possible to batch together all the above cal
...
+| `StopLoss`            | Compatible                              | Returns the same order whenever its checks pass and accepts empty `offchainInput`.                                        |
+| `TradeAboveThreshold` | Not supported by the current watchtower | `sellAmount` is the owner's balance, which funding changes before verification.                                             |
+| `PerpetualStableSwap` | Not supported by the current watchtower | The sell token and amounts depend on the owner's balances, which funding changes before verification.                       |
+| `GoodAfterTime`       | Not compatible                          | It gets `buyAmount` from the `offchainInput` passed to `getTradeableOrderWithSignature`, but the Poller passes empty bytes. |
+
+A modified watchtower could simulate funding first, then query balance-based handlers. `TradeAboveThreshold` cannot fund an empty owner.
```
anxolin: ok, I see u add here, then maybe is good in the table you hint the solution is described bellow, and should be clear the contract is actually compatible

src/types/ComposableCowPoller.sol
igorroncevic: resolved in [d25c032](https://github.com/cowprotocol/composable-cow/commit/d25c032e9bf42373ef78859ce2195f281f68d44c)

README.md
igorroncevic: resolved in [d25c032](https://github.com/cowprotocol/composable-cow/commit/d25c032e9bf42373ef78859ce2195f281f68d44c)

README.md
igorroncevic: resolved in [d25c032](https://github.com/cowprotocol/composable-cow/commit/d25c032e9bf42373ef78859ce2195f281f68d44c)

README.md
igorroncevic: resolved in [d25c032](https://github.com/cowprotocol/composable-cow/commit/d25c032e9bf42373ef78859ce2195f281f68d44c)

README.md
igorroncevic: resolved in [d25c032](https://github.com/cowprotocol/composable-cow/commit/d25c032e9bf42373ef78859ce2195f281f68d44c)

anxolin (approved):

